PR #82, opened 2026-02-03
Implement binary storage service

ginaxu1:
## Summary
Implemented a new binary storage service in the backend to handle file uploads and downloads. This service abstracts the storage backend, supporting both local filesystem storage (for development) and AWS S3 (for production), configured via environment variables

## Type of Change

## Changes Made
- added `backend/internal/uploads` package
- defined `StorageDriver` interface for pluggable storage backends
- implemented `LocalFSDriver` (with directory hashing) and S3Driver using AWS SDK v2
- added `UploadService` to manage file operations
- added HTTPHandler with `POST /upload` and `GET /download/:key` endpoints
- updated Config struct to include `StorageConfig` for flexible setup
- added `NewStorageFromConfig` to initialize the correct driver based on `STORAGE_TYPE`

## Testing
run `go test -v ./internal/uploads/...` in backend directory

## Deployment Notes
New environment variables are required:
- STORAGE_TYPE (defaults to "local", options: "local", "s3")
- STORAGE_LOCAL_BASE_DIR (if local)
- STORAGE_S3_BUCKET, STORAGE_S3_REGION, STORAGE_S3_ACCESS_KEY, STORAGE_S3_SECRET_KEY (if s3)


ginaxu1:
@sthanikan2000 let's keep `http_handler`, without it, the frontend has no way to get the file to the server before or during the form submission process unless we build a custom multipart handler for every domain endpoint

PR #140, opened 2026-02-12
feat: Update routers and services to support pagination

ginaxu1:
closes #118, **this is breaking change for trader-app** so see https://github.com/OpenNSW/nsw/pull/142 for trader-app updates. `oga-app` will continue to work with its existing logic since its backend hasn't switched to pagination yet

https://github.com/user-attachments/assets/6e943fc1-257d-46fd-abce-1ab665015c4f

## Summary
Add API pagination for the consignment (limit 50) and pre-consignment (limit 15) endpoints. now backend returns paginated response structure (totalCount, items, offset, limit) and optimizes data retrieval by performing pagination at the db level

## Type of Change
- [x] Performance improvement
- [x] Refactor

## Changes Made
  - refactored `PreConsignmentService` + `ConsignmentService` to use OFFSET and LIMIT in db queries. this replaces the previous in-memory pagination for pre-consignments, reducing memory overhead
- paginated response DTOs: updated response structures to include pagination metadata (totalCount, offset, and limit) to support frontend display logic (e.g. "Page 1 of 10")
- update OpenAPI specification to reflect pagination changes
- add `utils.ParsePaginationParams` to centralize the parsing and validation of offset and limit query parameters across all routers

## Testing
unit tests & manual testing done (omitted from this PR)

Script for manual testing below

```
#!/bin/bash
set -e

BASE_URL="http://localhost:8080/api/v1"
TRADER_ID="TRADER-001"
TEMPLATE_ID="493d31d3-bfb4-489f-94a0-5a3ca2e7ca01"

echo "1. Creating Pre-Consignment..."
CREATE_RES=$(curl -s -X POST "$BASE_URL/pre-consignments" \
  -H "Content-Type: application/json" \
  -d "{\"traderId\": \"$TRADER_ID\", \"preConsignmentTemplateId\": \"$TEMPLATE_ID\"}")

echo "Response: $CREATE_RES"
ID=$(echo $CREATE_RES | jq -r '.id')
TASK_ID=$(echo $CREATE_RES | jq -r '.workflowNodes[0].id')
echo "Created Pre-Consignment ID: $ID"
echo "Task ID: $TASK_ID"

if [ "$ID" == "null" ] || [ "$TASK_ID" == "null" ]; then
  echo "Failed to create pre-consignment or extract IDs"
  exit 1
fi

echo -e "\n2. Getting Pre-Consignment by ID..."
curl -s "$BASE_URL/pre-consignments/$ID" | jq .

[...]

PR #142, opened 2026-02-12
feat: Update frontend to support paginated API responses

### LINKED ISSUE #118: Add pagination support to consignments and pre-consignments list endpoints
### Current Limitation

The `GET /api/v1/consignments` and `GET /api/v1/pre-consignments` endpoints return all results without pagination support. This can cause performance issues and poor user experience when traders have large numbers of consignments or pre-consignments, as the entire list will be loaded at once without any ability to fetch results in chunks.

### Suggested Improvement

Add pagination parameters `offset` and `limit` to both endpoints:

1. GET /api/v1/consignments - Add optional query parameters:
    - `offset` (default: 0) - pagination offset
    - `limit` (default: 50) - number of items per page
    - Return a paginated response structure: `{ totalCount, items, offset, limit }`

2. GET /api/v1/pre-consignments - Add optional query parameters:
    - `offset` (default: 0) - pagination offset
    - `limit]` (default: 50) - number of items per page
    - Return a paginated response structure: `{ totalCount, items, offset, limit }`

Both endpoints should maintain the required `traderId` query parameter for filtering by trader.

### Version

_No response_

### Additional Context

- The `GET /api/v1/hscodes` endpoint already implements pagination successfully with `offset`, `limit]`, and returns `HSCodeListResult` with `totalCount`
- This follows the same pagination pattern used elsewhere in the system
- Pagination is essential for scalability, especially when dealing with large volumes of consignments in production
- The OpenAPI specification will need to be updated once this feature is implemented

ginaxu1:
https://github.com/user-attachments/assets/818a6824-cd75-44f4-afc4-e92cc221c769

https://github.com/user-attachments/assets/0353843f-8aa6-4d57-a6cb-324934c4f69d

closes #118, related #140
 
## Summary
Update `trader-app` to support the new paginated response structure from the backend API, handlning the transition from raw arrays to paginated objects `{ totalCount, items, offset, limit }` for both consignments and pre-consignments.

## Changes Made
- update `getAllConsignments` in `consignment.ts` to handle both legacy array responses and the new paginated object structure. This ensures the app doesn't crash during the backend transition
-  add `ConsignmentListResult` to the frontend type definitions to match the backend's `ConsignmentListResult` DTO
- update `ConsignmentScreen.tsx` to use the items and totalCount fields from the paginated API response instead of relying on array length

### DIFF
--- a/portals/apps/trader-app/src/screens/PreconsignmentScreen.tsx
+++ b/portals/apps/trader-app/src/screens/PreconsignmentScreen.tsx
@@ -29,13 +29,18 @@ export function PreconsignmentScreen() {
     const navigate = useNavigate()
     const [loading, setLoading] = useState(true)
     const [items, setItems] = useState<TraderPreConsignmentItem[]>([])
+    const [totalCount, setTotalCount] = useState(0)
+    const [page, setPage] = useState(0)
+    const PAGE_SIZE = 15
+
     const [notification, setNotification] = useState<{ type: 'success' | 'error', message: string } | null>(null)
 
     const loadData = async () => {
         try {
             setLoading(true)
-            const response = await getTraderPreConsignments()
+            const response = await getTraderPreConsignments(page * PAGE_SIZE, PAGE_SIZE)
             setItems(response.items || [])
+            setTotalCount(response.totalCount)
         } catch (error) {
             console.error('Failed to load pre-consignments', error)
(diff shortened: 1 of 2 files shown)


sthanikan2000 (changes requested):
I tested the frontend with merging the backend changes. I can see that these changes only getting the responses from the API endpoints in the correct DTO format. But not implemented the way to use the Pagination in the frontend

PR #176, opened 2026-02-23
feat: Conditional Workflow Unlocking and End-Node Completion

### LINKED ISSUE #174: [BE][WM] Conditional Workflow Unlocking and End-Node Completion
## Problem
Current workflow management unlocks a node only when all dependency nodes are completed, and marks a workflow as finished only when all nodes are completed. This strict behavior does not support real-world branching scenarios where:

- A node may become available when one of several dependency outcomes is met.
- Workflow completion may depend on reaching a specific end path rather than completing every possible node.

Because of this, workflows cannot model conditional progression cleanly (for example, approval/rejection branches), and completion logic can be overly restrictive.

## Proposed Solution
Introduce conditional unlock and end-node-based completion in workflow templates.

### 1) Workflow Node Template Enhancements
Add an optional UnlockConfiguration that supports boolean dependency logic using structured JSON (DNF style), while keeping DependsOn for dependency listing and backward compatibility.

- If UnlockConfiguration is not defined, keep current behavior (all DependsOn nodes must be completed).
- If UnlockConfiguration is defined, evaluate unlock conditions against dependency node completion outcomes.

### 2) Workflow Node Completion Outcomes
Keep node state as COMPLETED, and add a separate outcome/sub_state field to capture completion variants (for example: APPROVED, REJECTED, PARTIAL).

This allows unlock logic to target specific completion outcomes without expanding core state enums.

### 3) Workflow Template End Node
Add an optional EndNodeTemplateID to WorkflowTemplate.

- If EndNodeTemplateID is present, workflow is considered complete when that end node is completed (unlocked path reached and finished).
- If EndNodeTemplateID is absent, preserve existing all-nodes-completed logic.

### 4) Runtime Resolution
During workflow initialization:

- Resolve template-level dependency references to instance node IDs (existing pattern).
- Resolve UnlockConfiguration references similarly for runtime evaluation.

### 5) Compatibility and Migration
Implement as backward-compatible schema and logic updates:

- Existing templates continue working without changes.
- New fields are optional and nullable.
- Existing state-machine transitions remain intact, with outcome-aware completion checks added where needed.

## Alternatives
1. Keep strict all-dependencies/all-nodes logic and model branches as separate workflows.
   - Drawback: duplication, poor maintainability, weak traceability of business paths.

2. Use string-based expression parsing for unlock conditions.
   - Drawback: parser complexity, harder validation, and higher risk of runtime errors.

3. Encode outcomes as additional core states (for example COMPLETED_APPROVED).
   - Drawback: state explosion, schema churn, and more complex transition rules.

The proposed structured UnlockConfiguration + separate outcome field is the most maintainable and extensible approach.

sthanikan2000:
## Summary

This PR implements conditional workflow unlocking and end-node-based completion for Issue #174, and now supports generalized boolean unlock expressions.

Key capabilities:

1. **General unlock expressions**
   - Adds recursive boolean expression support with nested AND/OR logic via `expression`.
   - Keeps existing `anyOf/allOf` DNF payloads fully supported for backward compatibility.

2. **Completion outcomes**
   - Adds `outcome` on workflow nodes to represent completion sub-states (`APPROVED`, `REJECTED`, `FAST_TRACKED`, etc.) without expanding core state enums.

3. **End-node workflow completion**
   - Adds `EndNodeTemplateID` support so workflows can finish when a configured end node completes (instead of requiring all nodes to complete).

4. **Full backward compatibility**
   - Existing workflows without unlock configuration continue using legacy `depends_on` AND-all behavior.
   - Existing DNF unlock configuration continues to work unchanged.

## Type of Change

### Schema and Migrations

- Added `unlock_configuration JSONB` to `workflow_node_templates`.
- Added `outcome` and `unlock_configuration` to `workflow_nodes`.
- Added `end_node_template_id` to `workflow_templates`.
- Added `end_node_id` to `consignments` to track resolved end-node instance.

### Unlock Configuration Model

- Added generalized recursive expression support:
  - `expression.anyOf` for OR
  - `expression.allOf` for AND
  - leaf conditions with `nodeTemplateId` plus optional `state` and/or `outcome`
- Kept legacy `AnyOf/AllOf` DNF support in place.
- Validation now supports both modes and prevents ambiguous configs using both `expression` and legacy `anyOf` together.
- `ResolveToInstanceIDs` and `Evaluate` now support recursive expressions.

### Workflow/State Machine Behavior

- Unlock evaluation now supports generalized boolean expressions.
- `Outcome` is set on completion and used during unlock evaluation.
- End-node completion is supported via `WorkflowCompletionConfig`.
- Legacy behavior remains default when advanced config is absent.

### Seed Data

Updated Fresh Coconut seed workflow in `backend/internal/database/migrations/011_insert_unlock_config_seed.sql`:

- Node 1: unchanged (root, no unlock config)
- Node 2–5: migrated from DNF `anyOf/allOf` blocks to `expression`-based boolean configs
- Logic remains equivalent to previous behavior, but now demonstrates the generalized structure

## Testing

- Added/updated unit tests for:
  - legacy DNF validation and evaluation
  - recursive expression validation and evaluation
  - template-to-instance resolution for expression trees
  - JSON marshal/unmarshal for both legacy and expression formats
- Workflow state machine tests continue to pass with unlock + completion behavior.

## Backward Compatibility

- New fields are optional/nullable.
- Existing payloads and workflows continue to function:
  - no unlock config → legacy `depends_on` behavior
  - existing `anyOf/allOf` DNF config → still supported
  - new `expression` config → fully supported

## Related Issue

Closes #174

## Notes

- This PR now documents and demonstrates both unlock configuration styles:
  - Legacy DNF: `anyOf/allOf`
  - Generalized expression: `expression` with nested `anyOf/allOf`
- Recommended direction for new configs is `expression`, while preserving legacy compatibility for existing data.

### DIFF
--- a/backend/internal/workflow/model/unlock_config.go
+++ b/backend/internal/workflow/model/unlock_config.go
@@ -14,8 +14,10 @@ import (
 // Used within UnlockGroup to form AND conditions across multiple nodes.
 type UnlockCondition struct {
 	// NodeTemplateID references the workflow node template to check (template-level).
-	// At the instance level, this is resolved to the actual workflow node ID.
 	NodeTemplateID uuid.UUID `json:"nodeTemplateId"`
+
+	// NodeID is the resolved workflow node instance ID (after resolution). Optional in the condition definition, but will be populated during evaluation.
+	NodeID *uuid.UUID `json:"nodeId,omitempty"`
 
 	// State is the expected state of the referenced node (e.g., "COMPLETED", "FAILED").
 	// Optional — if nil, the node's state is not checked.
@@ -43,9 +45,10 @@ type UnlockExpression struct {
 	AnyOf []UnlockExpression `json:"anyOf,omitempty"`
 	AllOf []UnlockExpression `json:"allOf,omitempty"`
 
-	NodeTemplateID uuid.UUID `json:"nodeTemplateId,omitempty"`
-	State          *string   `json:"state,omitempty"`
-	Outcome        *string   `json:"outcome,omitempty"`
+	NodeTemplateID uuid.UUID  `json:"nodeTemplateId,omitempty"`
+	NodeID         *uuid.UUID `json:"nodeId,omitempty"`
+	State          *string    `json:"state,omitempty"`
+	Outcome        *string    `json:"outcome,omitempty"`
 }
 
 // UnlockConfig represents the unlock configuration for a workflow node.
@@ -124,6 +127,9 @@ func (uc *UnlockConfig) validateCondition(cond UnlockCondition, path string) err
 	if cond.NodeTemplateID == uuid.Nil {
 		return fmt.Errorf("%s has nil nodeTemplateId", path)
 	}
+	if cond.State == nil && cond.Outcome == nil {
+		return fmt.Errorf("%s must specify at least state or outcome", path)
+	}
 	if cond.State != nil && len(strings.TrimSpace(*cond.State)) == 0 {
 		return fmt.Errorf("%s has empty state", path)
 	}
@@ -184,6 +190,12 @@ func (uc *UnlockConfig) validateExpression(expr UnlockExpression, path string) e
 // ResolveToInstanceIDs creates a copy of the UnlockConfig with template IDs replaced by instance node IDs.
 // The templateToNodeID map should contain template ID -> node instance ID mappings.
 func (uc *UnlockConfig) ResolveToInstanceIDs(templateToNodeID map[uuid.UUID]uuid.UUID) (*UnlockConfig, error) {
+	// Validate the config before resolution
+	if err := uc.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid unlock configuration: %w", err)
+	}
+
+	// If it's an expression-based config, resolve the expression
 	if uc.Expression != nil {
 		resolvedExpr, err := uc.resolveExpressionToInstanceIDs(*uc.Expression, templateToNodeID)
 		if err != nil {
@@ -205,7 +217,8 @@ func (uc *UnlockConfig) ResolveToInstanceIDs(templateToNodeID map[uuid.UUID]uuid
 				return nil, fmt.Errorf("no instance node found for template ID %s in unlock configuration", cond.NodeTemplateID)
 			}
 			resolved.AnyOf[i].AllOf[j] = UnlockCondition{
-				NodeTemplateID: nodeID, // At instance level, this field holds the node instance ID
+				NodeTemplateID: cond.NodeTemplateID,
+				NodeID:         &nodeID,
 				State:          cond.State,
 				Outcome:        cond.Outcome,
 			}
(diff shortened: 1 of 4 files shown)


backend/internal/workflow/service/pre_consignment_service.go
```diff
@@ -408,6 +414,26 @@ func (s *PreConsignmentService) markPreConsignmentAsCompleted(ctx context.Contex
 	return nil
 }
 
+// getCompletionConfig builds a WorkflowCompletionConfig for a pre-consignment by looking up
+// the workflow template's EndNodeTemplateID via pre-consignment → pre-consignment template → workflow template.
+func (s *PreConsignmentService) getCompletionConfig(ctx context.Context, preConsignmentID uuid.UUID) (*WorkflowCompletionConfig, error) {
```
ginaxu1: `getCompletionConfig` missing in `consignment_service.go`.  if end-node completion should apply to consignments, add a `getCompletionConfig()` method (like you do for preconsignments here) and pass it to `TransitionToCompleted`.

backend/internal/workflow/service/consignment_service.go
ginaxu1: missing the optional `completionConfig`

backend/internal/workflow/service/pre_consignment_service.go
```diff
@@ -408,6 +414,26 @@ func (s *PreConsignmentService) markPreConsignmentAsCompleted(ctx context.Contex
...
+	var preConsignment model.PreConsignment
+	if err := s.db.WithContext(ctx).
```
ginaxu1: on every node completion, you're running 2 DB queries (that get the same data already available during init): 
1. `Preload("PreConsignmentTemplate").First(preConsignment)` to fetch pre-consignment + template
2. `GetWorkflowTemplateByID` to fetch the workflow template

suggest:
either cache the `EndNodeTemplateID` on the pre-consignment record or Pass the completion config into the service from the caller who already has it

backend/internal/task/manager/notification.go
```diff
@@ -11,4 +11,5 @@ type WorkflowManagerNotification struct {
 	UpdatedState        *plugin.State
 	AppendGlobalContext map[string]any
 	ExtendedState       *string
+	Outcome             *string // Outcome sub-state for COMPLETED transitions (e.g., APPROVED, REJECTED)
```
ginaxu1: why `Outcome` not in the `buildPreConsignmentResponseDTO`?  double-check all builders

sthanikan2000: Because currently we assumed that PreConsignments only have one workflow node

backend/internal/workflow/model/unlock_config.go
```diff
@@ -0,0 +1,160 @@
...
+func (uc *UnlockConfig) Validate() error {
+	if len(uc.AnyOf) == 0 {
+		return fmt.Errorf("unlock configuration must have at least one group in anyOf")
+	}
+	for i, group := range uc.AnyOf {
+		if len(group.AllOf) == 0 {
+			return fmt.Errorf("unlock configuration group %d must have at least one condition in allOf", i)
+		}
+		for j, cond := range group.AllOf {
+			if cond.NodeTemplateID == uuid.Nil {
```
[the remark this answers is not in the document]
sthanikan2000: When defining the Unlock Condition, it will be defined in the templates. We can't get a instance ID without creating instance for the nodes

PR #206, opened 2026-03-03
feat(upload): refactor Upload Service to improve security 

### LINKED ISSUE #199: Secure Upload Service
### Current Limitation

Current upload service in `backend/internal/uploads` is a basic implementation that lacks integrated authentication checks and robust protections against resource exhaustion (DDoS).

### Suggested Improvement

- [ ] add TTL for read/download file
- [ ] add WSO2 Thunder (JWT Verification)
- [ ] add `internal/auth/middleware.go`
- [ ] look into rate limiter: `golang.org/x/time/rate`
- [ ] add S3Storage with Pre-signed URL support

### Version

_No response_

### Additional Context

_No response_

### LINKED ISSUE #213: [FE][SIMPLE_FORM] File Renderer for field format `file` broken
### Describe the Bug

Even though we're getting the following response, when making the `GET` request for the `api/v1/tasks/:taskId`, which returning the `"format": "file"` for the certificate fields, UI not rendering the field as `file` format:
```json
{
    "success": true,
    "data": {
        "type": "SIMPLE_FORM",
        "pluginState": "INITIALIZED",
        "state": "INITIALIZED",
        "content": {
            "traderFormInfo": {
                "formData": {
                    ...
                },
                "schema": {
                    "type": "object",
                    "required": [
                        ...,
                        "tinCertificate",
                        "vatCertificate"
                    ],
                    "properties": {
                        ...,
                        "tinCertificate": {
                            "type": "string",
                            "title": "TIN Registration Certificate",
                            "format": "file",
                            "example": "tin_certificate.pdf",
                            "x-globalContext": {
                                "writeTo": "br:tinCertificate"
                            }
                        },
                        "vatCertificate": {
                            "type": "string",
                            "title": "VAT Registration Certificate",
                            "format": "file",
                            "example": "vat_certificate.pdf",
                            "x-globalContext": {
                                "writeTo": "br:vatCertificate"
                            }
                        },
                        ...
                    }
                },
                "title": "General Trader Verification",
                "uiSchema": {
                    "type": "VerticalLayout",
                    "elements": [
                        ...,
                        {
                            "type": "Control",
                            "scope": "#/properties/tinCertificate"
                        },
                        ...,
                        {
                            "type": "Control",
                            "scope": "#/properties/vatCertificate"
                        }
                    ]
                }
            }
        }
    }
}
```

### To Reproduce

1. Add a configuration for any form template with file format
2. Render that form during the workflow execution

### Expected Behavior

It should render the file same as the video in the PR description of https://github.com/OpenNSW/nsw/pull/148

### Screenshots

Image

### Version

main branch (eba5272c07c93c35b87e4a6d7edf520d718d7166)

### Environment Details (with versions)

_No response_

### Additional Context

Can refer to the `demo-branch` with the commit (1f9f580ff1e7d8a000c3f50fbb79014303ab2b71) to refer the previous implementation

ginaxu1:
closes #199 and closes #213

https://github.com/user-attachments/assets/da616662-9cb5-4de2-8fac-93ff620f46f5

## Summary
Refactor Upload Service to mitigate DDoS risks and enhance security by shifting from direct backend streaming to _TTL-based presigned URLs and authenticated identity checks_. UI now has a secure, authenticated upload service that returns a JSON response containing a time-limited download URL. Also added`StorageDriver` interface supporting S3 presigning and local storage with UUID-based naming and two-level directory hashing. `UploadService` coordinates file persistence and cleanup, while a `Factory` dynamically selects the storage driver based on env configs

#### Backend
**API contract**:
- `GET /api/v1/uploads/{key}` returns JSON `{"download_url": "...", "expires_at": <unix_timestamp>}`. For S3, download_url is a 15-minute presigned GET URL. For LocalFS, it points to `/api/v1/uploads/{key}/content`.
- `GET /api/v1/uploads/{key}/content` streams the file body with appropriate Content-Type, Content-Disposition, and Content-Length headers. Only functional when using `LocalFSDriver` (returns 404 otherwise). No auth required on this endpoint, it serves the same role as an S3 presigned URL; the caller was already authenticated when obtaining the URL via `GET /uploads/{key}`
- `POST /api/v1/uploads` upload with server-side MIME sniffing and content-type allowlist (application/pdf, image/jpeg, image/png, image/gif, image/webp). Returns file metadata with an empty url field. Clients should call `GET /uploads/{key}` when they need a download URL.
- `DELETE /api/v1/uploads/{key}` deletes the file from storage

**Authentication**: 
- `POST /api/v1/uploads`, `GET /api/v1/uploads/{key}`, and `DELETE /api/v1/uploads/{key}` require a valid JWT auth context. `GET /api/v1/uploads/{key}/content` does not require auth. Unauthenticated requests on protected endpoints receive 401 with a JSON error body
 
**Storage drivers**:
- `StorageDriver` interface: renamed GenerateURL to GetDownloadURL with a ttl parameter
- S3Driver: `GetDownloadURL` generates a presigned GET URL (default 15 minutes). Removed PublicURL fallback — always presigns. Added RequestChecksumCalculation option to support uploads over HTTP (e.g. local MinIO)
- LocalFSDriver: `GetDownloadURL` returns `{PublicURL}/{key}/content` (e.g. /api/v1/uploads/{key}/content). Added resolveAndValidate with symlink-aware path traversal protection and sentinel ErrInvalidPath for callers to check with errors.Is. Config default for STORAGE_LOCAL_PUBLIC_URL changed from /bucket to /api/v1/uploads.
- config.go: Added strings.TrimSpace on STORAGE_TYPE to prevent whitespace-induced mismatches.

 **Upload service**:
- countingReader / countingReadSeeker: wraps the upload stream to count actual bytes written. countingReadSeeker is only constructed when the underlying reader supports seeking, so S3 SDK type assertions to `io.ReadSeeker` get a truthful answer. Byte counter resets on seek-to-zero (retry support)
- Upload() now reports cr.n (actual bytes written) as Size instead of trusting the client-supplied Content-Length. URL field is left empty, decoupled from upload
 
#### Frontend
- New `UploadContext` (UploadProvider, useUpload): host apps inject onUpload, getDownloadUrl, and openFileInNewTab callbacks. The renderer never calls fetch directly
- `GetDownloadUrl` return type changed from `Promise<string>` to `Promise<DownloadUrlResult>` ({ url: string; expiresAt: number }), so the backend-provided TTL flows through to the download URL cache instead of hardcoding 15 minutes
- FileControl: rewired from inline `FileReader.readAsDataURL` to the onUpload callback. Stores a `localBlobUrl` from the uploaded File object so the View button can open it instantly without a server round-trip. localBlobUrl being non-null distinguishes "just uploaded" from "pre-existing". Input value cleared on upload failure so the user can retry with the same file
- FileControlTester: now matches both data-url and file format
- Bounded download URL cache with configurable TTL buffer and max size (50 entries)

**trader-app and oga-app**:
- Added `UploadWrapper` component in App.tsx wiring UploadProvider with app-specific callbacks
- `services/upload.ts` uploadFile now accepts ApiClient and sends auth headers (without Content-Type: application/json to avoid corrupting multipart/form-data). getDownloadUrl returns `{ url, expiresAt }` from the backend response. openFileInNewTab changed from directly fetching /uploads/{key}/content to calling getDownloadUrl (hits GET /uploads/{key}) and then window.open(url) — relative URLs are resolved against the backend origin to prevent the SPA from intercepting them
- `trader-app/services/api.ts` added getAuthHeaders method to ApiClient interface with includeJsonContentType flag

## Testing
- verify `401 Unauthorized` responses on anonymous download access
- confirm valid TRADER JWTs receive the expected JSON redirect contract
- verify unit tests pass via `go test ./... -count=1`

### How Files are Stored
The StorageDriver interface manages files through Save, Get, Delete, GenerateURL, and GetDownloadURL. The service layer maps upload/download to Save/Get and uses GetDownloadURL (and, for S3, presigning) for secure download URLs. 
- system uses **UUIDs** (via the `google/uuid` package) for file naming to prevent collisions and ensure data integrity
- for local storage, the system implements **directory hashing**: a 2-level subdirectory layout based on the storage key (the first two characters, then the next two, then the full key), e.g. ab/cd/abcd1234-....pdf, so files are not all in one flat directory and filesystem performance is preserved
- the `UploadService` coordinates the upload flow and metadata: it generates the UUID-based key, calls the driver’s Save, then GenerateURL (and cleans up with Delete if URL generation fails), and returns FileMetadata

**Where Files are Uploaded**
The destination is determined by the storage driver selected at runtime via the factory (STORAGE_TYPE):
* **Local**: files are stored on disk in a configurable base directory. Directory hashing is applied under this base. Used for local development or simple deployments.
* **S3-compatible**: for production, the system supports S3-compatible backends, including AWS S3. The integration uses the AWS SDK for Go v2; when `STORAGE_TYPE=s3`, the factory creates an S3 client (optional custom endpoint and static credentials). Presigned GET URLs are used for secure, time-limited downloads; AWS S3 is the default when no custom endpoint is set

### DIFF
--- a/portals/apps/oga-app/src/services/upload.ts
+++ b/portals/apps/oga-app/src/services/upload.ts
@@ -41,24 +41,8 @@ export async function getDownloadUrl(apiClient: ApiClient, key: string): Promise
   }
 
   const data = (await response.json()) as { download_url: string; expires_at: number }
-  return { url: data.download_url, expiresAt: data.expires_at }
-}
-
-/** Fetch file content with auth and open in a new tab (for View button). Avoids relative download_url opening on frontend origin. */
-export async function openFileInNewTab(apiClient: ApiClient, key: string): Promise<void> {
-  const response = await fetch(`${API_BASE_URL}/uploads/${key}/content`, {
-    headers: await apiClient.getAuthHeaders(false),
-  })
-  if (!response.ok) {
-    throw new Error(`Failed to open file: ${response.status}`)
-  }
-  const blob = await response.blob()
-  const url = URL.createObjectURL(blob)
-  const w = window.open(url, '_blank', 'noopener,noreferrer')
-  if (w) {
-    setTimeout(() => URL.revokeObjectURL(url), 60_000)
-  } else {
-    URL.revokeObjectURL(url)
-    throw new Error('Popup blocked')
-  }
+  const url = data.download_url.startsWith('/')
+    ? `${new URL(API_BASE_URL).origin}${data.download_url}`
+    : data.download_url
+  return { url, expiresAt: data.expires_at }
 }
--- a/portals/packages/jsonforms-renderers/src/renderers/FileControl.tsx
+++ b/portals/packages/jsonforms-renderers/src/renderers/FileControl.tsx
@@ -50,7 +50,6 @@ const FileControl = ({ data, handleChange, path, label, required, uischema, enab
     const [downloadUrl, setDownloadUrl] = useState<string | null>(null);
     const [downloadLoading, setDownloadLoading] = useState(false);
     const [downloadError, setDownloadError] = useState<string | null>(null);
-    const [viewOpening, setViewOpening] = useState(false);
     const [localBlobUrl, setLocalBlobUrl] = useState<string | null>(null);
     const inputRef = useRef<HTMLInputElement>(null);
 
@@ -258,23 +257,6 @@ const FileControl = ({ data, handleChange, path, label, required, uischema, enab
                                 <Text size="1" color="gray">Loading...</Text>
                             ) : downloadError ? (
                                 <Text size="1" color="red">Error</Text>
-                            ) : viewOpening ? (
-                                <Text size="1" color="gray">Opening…</Text>
-                            ) : !localBlobUrl && uploadContext?.openFileInNewTab && data && isFileKey(data) ? (
-                                <Button
-                                    variant="soft"
-                                    color="blue"
-                                    size="1"
-                                    onClick={(e) => {
-                                        e.stopPropagation();
-                                        const open = uploadContext.openFileInNewTab;
-                                        if (!open) return;
-                                        setViewOpening(true);
-                                        open(data).finally(() => setViewOpening(false));
-                                    }}
-                                >
-                                    View
-                                </Button>
                             ) : (
                                 <Button variant="soft" color="blue" size="1" asChild>
                                     <a
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -112,7 +112,7 @@ func Load() (*Config, error) {
 		Storage: StorageConfig{
 			Type:           strings.TrimSpace(getEnvOrDefault("STORAGE_TYPE", "local")),
 			LocalBaseDir:   getEnvOrDefault("STORAGE_LOCAL_BASE_DIR", "./bucket"),
-			LocalPublicURL: getEnvOrDefault("STORAGE_LOCAL_PUBLIC_URL", "/bucket"),
+			LocalPublicURL: getEnvOrDefault("STORAGE_LOCAL_PUBLIC_URL", "/api/v1/uploads"),
 			S3Endpoint:     os.Getenv("STORAGE_S3_ENDPOINT"),
 			S3Bucket:       getEnvOrDefault("STORAGE_S3_BUCKET", "nsw-uploads"),
 			S3Region:       getEnvOrDefault("STORAGE_S3_REGION", "us-east-1"),
--- a/backend/internal/uploads/drivers/local_fs.go
+++ b/backend/internal/uploads/drivers/local_fs.go
@@ -136,10 +136,8 @@ func (d *LocalFSDriver) Delete(ctx context.Context, key string) error {
 }
 
 func (d *LocalFSDriver) GetDownloadURL(ctx context.Context, key string, ttl time.Duration) (string, error) {
-	// For local development, returning the standard URL is acceptable.
-	// The frontend will treat this the same as an S3 presigned URL.
 	if d.PublicURL == "" {
 		return key, nil
 	}
-	return fmt.Sprintf("%s/%s", d.PublicURL, key), nil
+	return fmt.Sprintf("%s/%s/content", d.PublicURL, key), nil
 }
--- a/backend/internal/uploads/drivers/local_fs_test.go
+++ b/backend/internal/uploads/drivers/local_fs_test.go
@@ -58,7 +58,8 @@ func TestLocalFSDriver_DirectoryHashing(t *testing.T) {
 	if err != nil {
 		t.Errorf("GetDownloadURL failed: %v", err)
 	}
-	if !strings.HasSuffix(url, key) || !strings.Contains(url, "/uploads") {
+	expectedSuffix := key + "/content"
+	if !strings.HasSuffix(url, expectedSuffix) || !strings.Contains(url, "/uploads") {
 		t.Errorf("unexpected URL: %s", url)
 	}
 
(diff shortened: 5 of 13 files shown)


backend/cmd/server/main.go
```diff
@@ -131,6 +131,7 @@ func main() {
 
 	// Upload routes
 	mux.HandleFunc("POST /api/v1/uploads", uploadHandler.Upload)
+	mux.HandleFunc("GET /api/v1/uploads/{key}/content", uploadHandler.DownloadContent)
```
mushrafmim: Why this endpoint needed? We are providing a presigned URL to the user.

portals/packages/jsonforms-renderers/src/renderers/FileControl.tsx
```diff
@@ -43,33 +149,31 @@ const FileControl = ({ data, handleChange, path, label, required, uischema, enab
...
+            try {
+                const result = await uploadContext.onUpload(file);
+                handleChange(path, result.key);
+                setFileName(result.name ?? file.name);
+                setError(null);
+            } catch {
+                setError('Upload failed.');
+            }
+            return;
+        }
```
mushrafmim: After a successful upload the local File object is discarded and data is set to the key. This immediately triggers fetchDownloadUrl (via the useEffect on line 94) and the View button falls through to either openFileInNewTab or the server-fetched downloadUrl — both of which hit the server. Since the binary is     
still in memory at this point, a blob URL should be created from the File before it's discarded and stored in a separate local state (e.g. localBlobUrl), so the View button can open it without a round-trip until the component unmounts.

backend/internal/uploads/service.go
```diff
@@ -15,6 +16,17 @@ type UploadService struct {
...
+	r io.Reader
+	n int64
+}
+
+func (c *countingReader) Read(p []byte) (int, error) {
+	n, err := c.r.Read(p)
+	c.n += int64(n)
+	return n, err
+}
+
```
mushrafmim: medium

Seek forwards to the underlying reader if it supports seeking, preserving transparency so callers (e.g. the S3 SDK) can still type-assert to io.ReadSeeker. The byte count is reset when the stream is rewound to the beginning.

```suggestion
func (c *countingReader) Seek(offset int64, whence int) (int64, error) {
	seeker, ok := c.r.(io.Seeker)
	if !ok {
		return 0, fmt.Errorf("underlying reader does not support seeking")
	}
	pos, err := seeker.Seek(offset, whence)
	if err == nil && pos == 0 {
		c.n = 0
	}
	return pos, err
}
```

portals/apps/oga-app/src/services/upload.ts
```diff
@@ -28,10 +27,38 @@ export async function uploadFile(apiClient: ApiClient, file: File): Promise<File
...
+  const blob = await response.blob()
+  const url = URL.createObjectURL(blob)
+  const w = window.open(url, '_blank', 'noopener,noreferrer')
+  if (w) {
+    setTimeout(() => URL.revokeObjectURL(url), 60_000)
+  } else {
+    URL.revokeObjectURL(url)
+    throw new Error('Popup blocked')
+  }
 }
```
mushrafmim: openFileInNewTab calls `GET /uploads/{key}/content` to proxy the file through the backend, but this is unnecessary. The correct behaviour is to call `GET /uploads/{key}` to obtain the presigned URL and open that directly in a new tab — which is exactly what getDownloadUrl already does. The `/content` endpoint should
  not be used here; instead, openFileInNewTab should be removed and the existing resolvedHref (presigned URL) path in FileControl.tsx should handle the View button for externally-sourced file keys.

sthanikan2000 (dismissed):
I rebased your branch with `OpenNSW/nsw:main` and I tried the end to end flow
Trader App side changes working properly
But can't able to even get the login page for the OGA portal

So I tried the following commands and build failing:
Image

Issue #198, opened 2026-03-03
Add Select CHA to workflow


ginaxu1:
### Problem
Currently our system does not support new entity: Clearing House Agent (CHA). 

### Proposed Solution

1st step in the Trader workflow should be to "Select a CHA"

CHA Portal will be similar to Trader Portal except

- [ ] No "New Consignment" button in /consignments
- [ ] No Verified docs sidebar tab
- [ ] Restrict actions to CHA vs Trader, difference between Trader and CHA: there are certain nodes only trader can edit (CHA can't edit) and vice versa

PR #234, opened 2026-03-09
feat(cha): add backend updates for CHA entity and create Consignment 

ginaxu1:
Related to #198 

## Note 
Verified manually these updates **not** breaking existing implementation. Legacy support for single-stage creation is preserved for flows where the trader manages the entire process: 

- `POST /api/v1/consignments`- if the body is { "flow", "items" } (no chaId), the code still calls the original `InitializeConsignment` and creates the consignment + workflow in one step. So the existing “create with HS code” flow is unchanged.
- `GET /api/v1/consignments`- if the request has no role (or only pagination/filters), we default role=trader and use the auth context trader_id as before. So existing list calls keep working; we only added support for `?role=cha&cha_id=....`
- `GET /api/v1/consignments/{id}` unchanged; we only added an optional `chaId` on the response. Old clients can ignore it
- Response DTOs: added optional `chaId` to `detail/summary`
- `PUT /api/v1/consignments/{id}/initialize`: new endpoint; existing clients don’t call it, so no impact on current behavior.

## Summary
Add support for Customs House Agents (CHAs) with a 2-stage consignment initiation process. This allows Traders to select a CHA service provider to manage their consignment before CHA selects a specific HS code

## Type of Change
[x] New feature (non-breaking change which adds functionality)

## Changes 
* Create `CHA` model and implemented a database migration to create the `customs_house_agents` table. The migration also seeds five major service providers
**Two-Stage Consignment Initiation**:
* 1: add `CreateConsignmentShell` in the service layer, allowing a Trader to create a consignment shell assigned to a specific `cha_id` with a new `AWAITING_INITIATION` state
* 2: add `InitializeConsignmentByID` to allow a CHA to select an HS code and trigger the BPMN workflow creation, transitioning the state to `IN_PROGRESS`
* Schema updates: add `cha_id` to the `consignments` table and updated state constraints to support `AWAITING_INITIATION`
* Update `ListConsignments` and relevant router handlers to support filtering by `role=cha` and `cha_id`, enabling agents to view only the consignments assigned to them.

**API Routing**:
* Update `POST /api/v1/consignments` to detect Stage 1 requests (containing a `chaId`) while maintaining backward compatibility for legacy single-stage initiations.
* Add a new endpoint `PUT /api/v1/consignments/{id}/initialize` for Stage 2 processing.

### DIFF
--- a/backend/internal/workflow/router/consignment_router.go
+++ b/backend/internal/workflow/router/consignment_router.go
@@ -2,6 +2,7 @@ package router
 
 import (
 	"encoding/json"
+	"log/slog"
 	"net/http"
 
 	"github.com/google/uuid"
@@ -54,7 +55,9 @@ func (c *ConsignmentRouter) HandleCreateConsignment(w http.ResponseWriter, r *ht
 		}
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusCreated)
-		_ = json.NewEncoder(w).Encode(consignment)
+		if err := json.NewEncoder(w).Encode(consignment); err != nil {
+			slog.Error("failed to encode response for consignment", "error", err)
+		}
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


ginaxu1:
Created #236

PR #275, opened 2026-03-23
Update to Flexible Storage Keys and fix OGA File Rendering

### LINKED ISSUE #274: Implement Upload/Download Suggestions
### Suggested Improvement
1. Update the storage key validation in `backend/internal/uploads/http_handler.go:L18` to allow more diverse key formats (e.g. descriptive names, subdirectories) while ensuring security against path traversal and other malicious keys

2. https://github.com/OpenNSW/nsw/pull/205#issuecomment-4108074496

### Version

_No response_

### Additional Context

_No response_

ginaxu1:
Closes #274 

## Testing
* verified that opening a review screen with multiple file fields no longer spams the `/oga/uploads` endpoint. Run `./start-dev.sh` and go to OGA app like FCAU `http://localhost:5175/`, verify API calls when click an application and click "View" next to an uploaded file . 
* Verified that clicking "View" successfully opens the document in a new tab without prompting for re-authentication
* Unit tests updated and passing (new tests included to test flexible storage keys)

## Summary
Address file rendering issues on the review screens where multiple unintended API calls were being made on mount, and clicking "View" forced the user to log in again in a new tab. Also improve the backend upload handlers to support flexible, nested storage keys with wildcard routing.

#### Frontend (JSONForms Renderers & Apps)
* Replace `getDownloadUrl` with a new `viewFile` handler in `UploadContext.tsx`
* Remove `useEffect` fetch logic in `FileControl.tsx`. Files are now only fetched when the user explicitly clicks the "View" button, resolving the API spam issue on screen load
* Implement the `viewFile` service in the app layer (`upload.ts`). It now fetches the actual file securely by attaching the authenticated `Bearer` token via the API client, converts the response to a `Blob`, and generates a local `ObjectURL`
* Bypass Re-login Prompt: Update the `FileControl` view button to open the generated `ObjectURL` in a new tab. Since it's a local memory reference, it bypasses the need for the new tab to pass through the API's auth middleware

#### Backend (Uploads Service)
* Flexible storage keys: update the `storageKeyRx` regex validation in `http_handler.go` to support alphanumeric characters, dashes, underscores, dots, and slashes safely (`^[a-zA-Z0-9-_./]+$`)
* Wildcard routing & multiplexer fix: Refactor the content download route to avoid a panic with Go 1.22's native multiplexer (which requires wildcards to be at the end of the pattern and can conflict with sub-routes)
  * Changed the route from `/api/uploads/content/{key...}` to `/api/uploads/download/{key...}`
  * Update `LocalFSDriver` to generate download URLs using this new prefix, verified compatible with the frontend.
* S3 Presign Fix: fix the `PresignGetObject` call in `s3_driver.go` to properly handle and return errors during URL generation

### DIFF
--- a/backend/internal/auth/auth_test.go
+++ b/backend/internal/auth/auth_test.go
@@ -40,7 +40,7 @@ func TestTokenExtractor_ExtractClaimsFromHeader(t *testing.T) {
 	}))
 	defer jwksServer.Close()
 
-	extractor, err := NewTokenExtractor(jwksServer.URL, "https://localhost:8090/oauth2/token", "TRADER_PORTAL_APP", "TRADER_PORTAL_APP")
+	extractor, err := NewTokenExtractor(jwksServer.URL, "https://localhost:8090/oauth2/token", []string{"TRADER_PORTAL_APP"}, []string{"TRADER_PORTAL_APP"})
 	if err != nil {
 		t.Fatalf("failed to create token extractor: %v", err)
 	}
@@ -342,7 +342,7 @@ func BenchmarkTokenExtraction(b *testing.B) {
 		b.Fatalf("failed to sign token: %v", err)
 	}
 
-	extractor, err := NewTokenExtractor(jwksServer.URL, "https://localhost:8090/oauth2/token", "TRADER_PORTAL_APP", "TRADER_PORTAL_APP")
+	extractor, err := NewTokenExtractor(jwksServer.URL, "https://localhost:8090/oauth2/token", []string{"TRADER_PORTAL_APP"}, []string{"TRADER_PORTAL_APP"})
 	if err != nil {
 		b.Fatalf("failed to create token extractor: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


backend/internal/app/bootstrap/app.go
```diff
@@ -157,14 +156,9 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
...
-	mux.Handle("DELETE /api/v1/uploads/{key}", withAuth(http.HandlerFunc(uploadHandler.Delete)))
-
-	// When using local storage, this endpoint serves the actual file bytes.
-	// It's made public since it's the equivalent of a presigned URL when using S3.
-	if _, ok := storageDriver.(*drivers.LocalFSDriver); ok {
-		mux.HandleFunc("GET /api/v1/uploads/{key}/content", uploadHandler.DownloadContent)
-	}
+	mux.Handle("GET /api/v1/contents/{key...}", withAuth(http.HandlerFunc(uploadHandler.DownloadContent)))
+	mux.Handle("GET /api/v1/uploads/{key...}", withAuth(http.HandlerFunc(uploadHandler.Download)))
+	mux.Handle("DELETE /api/v1/uploads/{key...}", withAuth(http.HandlerFunc(uploadHandler.Delete)))
```
mushrafmim: We don't need to change these endpoints. one call should support only one file at a time.

Issue #11, opened 2026-01-19
[Security] RBAC Middleware Implementation


ginaxu1:
💡 New Feature Request: Implement RBAC Middleware for NSW User Roles 

**Problem**: different users require distinct permission levels (e.g. a Trader should not access the Admin Portal, and an OGA Officer should only see tasks relevant to their agency)

**Proposed Solution**: Implement Role-Based Access Control (RBAC) in internal/api/middleware

- [ ] Define the role constants: ROLE_TRADER, ROLE_OGA_OFFICER, and ROLE_NSW_ADMIN
- [ ] Implement a middleware that extracts roles from the Asgardeo JWT claims
- [ ] Create a "Scope Guard" to protect specific routes

Example: `router.Use(auth.RequireRole(ROLE_NSW_ADMIN))` for workflow management endpoints

PR #296, opened 2026-03-31
RBAC for Endpoints (Trader/CHA) & M2M allowance for Tasks/Upload

ginaxu1:
Part 1/N of #11 
## Summary
Implements Phase 1 of a configurable RBAC system and adds support for M2M (Machine-to-Machine) authentication for OGA integrations. The backend has transitioned from hardcoded role assumptions to a dynamic, handler-level authorization model. Key architectural changes include making user identity optional to support service-to-service communication and moving permission boundaries to the entry points (HTTP handlers) rather than the business logic layer.

## Changes
* Defined fallback defaults in `backend/internal/config/config.go` 
```
InternalClientID:      getEnvOrDefault("AUTH_INTERNAL_CLIENT_ID", "OGA_M2M_APP")
TraderGroup:           getEnvOrDefault("AUTH_TRADER_GROUP", "NSW_TRADER")
CHAGroup:              getEnvOrDefault("AUTH_CHA_GROUP", "NSW_CHA")
```

### Authentication Core
* `backend/internal/auth/context.go`: refactor `AuthContext.UserID` to `*string` to support tokens without a subject (M2M/Client Credentials). add `ClientID`, `Groups []string`, and `IsM2M bool` fields to the context struct. introduce the `HasGroup(group string) bool` helper method for permission checking.
* `backend/internal/auth/token_parser.go`: enhance JWT parsing via JWKS to extract the `groups` claim. Add logic to identify M2M tokens specifically when the token is issued by the internal client and the subject matches the client ID or follows the M2M subject pattern.
* `backend/internal/auth/middleware.go`: update middleware to act as the primary enforcement gate, building the enriched `AuthContext`. It is now designed to gracefully tolerate users without a stored DB context (passing a `nil` UserContext) and skips mandatory database lookups if the token is identified as M2M.
* `backend/internal/config/config.go`: introduce configurable IDP group names (`TraderGroup`, `CHAGroup`) and an `InternalClientID` within the `AuthConfig` struct, mapping them to the corresponding environment variables.

### Authorization & Routing (Handlers)
* `backend/internal/workflow/router/consignment_router.go`: move role and identity validation to the HTTP handlers. Enforced human-only context for consignment creation and retrieval. update `HandleGetConsignments` to support a `role` query parameter, validating it against the user's actual IDP groups via `HasGroup()`.
* `backend/internal/task/manager/http_handler.go`: implement strict RBAC checks distinguishing between view and execute actions. For human users, both Traders and CHAs can view tasks (`HandleGetTask`). However, only CHAs (and M2M systems) can access the task execution endpoint (`HandleExecuteTask`). Traders are explicitly forbidden from executing tasks.
* `backend/internal/uploads/http_handler.go`: apply M2M bypass and human-role checking logic. Ensures that the `Upload`, `Download`, and `Delete` endpoints strictly require the requestor to possess either the Trader or CHA group, or be an authenticated M2M system account.

### Service Layer (Business Logic)
* `backend/internal/workflow/service/consignment_service.go`: include robust data-level RBAC filtering. The `GetConsignments` method dynamically applies database `WHERE` clauses (`trader_id = ?` vs. `cha_id = ?`) based on the explicit `userID` and `requestedRole` parameters passed down from the handlers, ensuring users only retrieve consignments they explicitly own or manage

## Validation/Testing
- Successfully executed full backend test suite (`go test ./...`) with all tests passing.
- Verified human user context: `GET /consignments?role=trader` with a `CHA` user correctly returns `403 Forbidden`.
- Verified M2M Block Test: `GET /consignments` using an M2M token correctly returns `401/403`.
- Verified M2M Bypass Test: `PATCH /tasks/:id` or upload endpoints using an M2M token correctly bypass human checks and return `200 OK`

### DIFF
--- a/backend/internal/app/bootstrap/app.go
+++ b/backend/internal/app/bootstrap/app.go
@@ -162,7 +162,7 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 		}
 
 		consignmentService := service.NewConsignmentService(db, templateService, nil, wmV2)
-		consignmentRouter = router.NewConsignmentRouter(consignmentService, chaService)
+		consignmentRouter = router.NewConsignmentRouter(consignmentService, chaService, &cfg.Auth)
 
 		// TODO: Pre-Consignment is commented out in new workflow for now
 		// preConsignmentService := service.NewPreConsignmentService(db, templateService, wmV2)
@@ -183,7 +183,7 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 		consignmentService := service.NewConsignmentService(db, templateService, wm, nil)
 		preConsignmentService := service.NewPreConsignmentService(db, templateService, wm)
 
-		consignmentRouter = router.NewConsignmentRouter(consignmentService, chaService)
+		consignmentRouter = router.NewConsignmentRouter(consignmentService, chaService, &cfg.Auth)
 		preConsignmentRouter = router.NewPreConsignmentRouter(preConsignmentService)
 	}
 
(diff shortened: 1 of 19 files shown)


sthanikan2000:
@ginaxu1  Don't waste time on RBAC for now. I'll bring the M2M Authentication first. Then it will be easier for you to implement RBAC authorization

ginaxu1:
Work done in https://github.com/OpenNSW/nsw/pull/361

PR #312, opened 2026-04-04
[Trader App] Derive roles from authenticated group claims and add configurable IdP group mapping

### LINKED ISSUE #309: [Trader App] Configure `availableGroups` from access token claims
## Current Limitation

In the `Trader App`, the available groups are currently hardcoded at https://github.com/OpenNSW/nsw/blob/321bca9fad65025dd2708a6b6dd77414a741bb6e/portals/apps/trader-app/src/App.tsx#L36

This means the app cannot adapt the allowed groups dynamically from the authenticated user context. If group membership changes, or if a different environment needs different groups, the code must be changed and redeployed.

## Suggested Improvement

Make `availableGroups` configurable from the authenticated user context, ideally by decoding the access token or by using another supported field from the `@asgardeo/react` package.

The goal is to derive the set of allowed groups from the signed-in user session instead of hardcoding them in the UI.

## Why This Helps

- Removes environment-specific hardcoding from the app.
- Keeps group configuration aligned with the identity provider.
- Makes it easier to support different roles or groups without rebuilding the frontend.
- Reduces the risk of stale authorization configuration in the client.

## Possible Approaches

1. Decode the access token and read group or role claims from the token payload.
2. Use a user info or claims API exposed by [Asgardeo React](https://wso2.com/asgardeo/docs/sdks/react/overview/) if it provides a more reliable source than raw token decoding.
3. Move the allowed group list to a runtime configuration source and populate it after authentication.

## Acceptance Criteria

- `availableGroups` is no longer hardcoded in `App.tsx`.
- The app reads the group list from authenticated user data or token claims.
- The solution works for the existing `trader` and `cha` groups.
- The UI still blocks access when the user is not signed in.

## Additional Context

This request is specifically about the `RoleProvider` usage in the trader app and how group availability should be sourced from authentication rather than embedded in the frontend code.

sthanikan2000:
## Summary

This PR removes hardcoded role availability in Trader App and derives roles from authenticated group claims.

It also adds centralized, configurable IdP group name mapping and updates local and Docker startup wiring so group-name overrides can be provided through environment variables.

Additionally, this PR includes the IdP sample resource update to ensure groups are present in id_token user attributes, so the frontend can read groups claim reliably.

## Type of Change

- [ ] Documentation update

## Changes Made

- Replaced hardcoded role list in Trader App with claim-driven role resolution from authenticated claims.
- Added exact-match claim-to-role mapping logic (Traders -> trader, CHA -> cha) with support for both roles when both groups are present.
- Moved pure claim-to-role mapping logic to utils for clearer separation of concerns.
- Extracted role resolution flow into a dedicated hook to reduce logic in App.tsx.
- Updated no-role experience to render via Layout shell in top-bar-only mode, with user dropdown available for sign-out.
- Kept no-role text exactly as: You are not assigned with any roles.
- Added runtime env configuration support for Trader and CHA IdP group names with defaults.
- Propagated new runtime env variables through:
  - trader app runtime env generation
  - local non-Docker startup script
  - root .env.example
  - root .env.docker.example
  - docker-compose trader-portal service environment
- Included IdP bootstrap/resource update so id_token user_attributes includes groups claim for frontend role derivation.

## Testing

- [x] All existing tests pass

### Validation performed

- trader-app type check passed
  - pnpm --filter trader-app type-check
- trader-app production build passed
  - pnpm --filter trader-app build

### Edge cases verified

- User with Traders only -> trader role
- User with CHA only -> cha role
- User with both groups -> both roles
- User with no recognized groups -> top-bar-only shell with user dropdown and no-role message

## Checklist

- [x] My changes generate no new warnings in changed scope

## Related Issues

Closes #309

## Screenshots/Demo

https://github.com/user-attachments/assets/7b3dfc28-ed63-4d40-af00-86b51fe7f035

## Additional Notes

Claim contract followed by this PR:

- Source: authenticated claims
- Claim: groups (string array)
- Matching: exact string matching only
- Default configured IdP group names:
  - Traders
  - CHA

## Deployment Notes

New optional environment variables for Trader App role mapping:

- TRADER_IDP_TRADER_GROUP_NAME (default: Traders)
- TRADER_IDP_CHA_GROUP_NAME (default: CHA)

These are now wired for both local and Docker startup flows.


ginaxu1 (dismissed):

ginaxu1 (changes requested):

ginaxu1 (dismissed):
Lgtm

Aravinda-HWK (dismissed):
LGTM

portals/apps/trader-app/src/components/Layout/Layout.tsx
mushrafmim, line 1: Do we need to complicate the `Layout.tsx` just to remove the `sideBar` for some person who don't have access to this system? Instead Can't we make the `NoRolesAssigned.tsx` as a screen and place the logout button to the application inside that. Such that we can render that screen (can call it UnAuthorizedScreen.tsx) outside of the `Layout` similar to how we are rendering the `LoginScreen`?

mushrafmim (changes requested):

portals/apps/trader-app/src/App.tsx
```diff
@@ -27,17 +29,24 @@ function UploadWrapper({ children }: { children: ReactNode }) {
 
 function ProtectedLayout() {
   const {isSignedIn, isLoading} = useAsgardeo()
+  const {availableRoles, isResolvingRoles} = useResolvedRoles(isSignedIn)
 
-  if (isLoading) return null
+  if (isLoading || (isSignedIn && (isResolvingRoles || availableRoles === null))) return null
```
mushrafmim: having another hook and passing the `isSignedIn` inside that hook complicates the logics, definitely this can be simplified.

mushrafmim (changes requested):

portals/apps/trader-app/src/components/Layout/Layout.tsx
sthanikan2000: Yeah better Idea

portals/apps/trader-app/src/App.tsx
sthanikan2000: will check seem I missed this one

sthanikan2000:
@mushrafmim I refactored my previous implementation according to the changes. Also I moved the implementations that was not related to `TopBar.tsx` to different files.
You can review it now.

portals/apps/trader-app/src/screens/UnauthorizedScreen.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Button } from '@radix-ui/themes'
+import { UserOnlyTopBar } from '../components/Layout/TopBar'
+import { useSignOutHandler } from '../hooks/useSignOutHandler'
+
+export function UnauthorizedScreen() {
+  const handleSignOut = useSignOutHandler()
+
+  return (
+    <div className="min-h-screen bg-gray-50">
+            <UserOnlyTopBar />
```
mushrafmim:
```suggestion

```
Why do we need this? We already have the Logout button in the middle of the screen. Can't we use the `Roles` hook and hide or show whatever necessary based on that? instead of introducing another piece of code.

portals/apps/trader-app/src/runtimeConfig.ts
```diff
@@ -2,6 +2,14 @@ type RuntimeConfigValue = string | undefined
 
 type RuntimeConfigMap = Record<string, RuntimeConfigValue>
 
+export interface IdpRoleGroupConfig {
+  traderGroupName: string
+  chaGroupName: string
+}
+
+const DEFAULT_TRADER_GROUP_NAME = 'Traders'
+const DEFAULT_CHA_GROUP_NAME = 'CHA'
```
mushrafmim, line 11: Why shouldn't we always use `lowercase` values for Groups, which will reduce the complicated logic of lowercasing and uppercasing, reduce overhead of exactly knowing what case it is?

portals/apps/trader-app/src/runtimeConfig.ts
```diff
@@ -2,6 +2,14 @@ type RuntimeConfigValue = string | undefined
 
 type RuntimeConfigMap = Record<string, RuntimeConfigValue>
 
+export interface IdpRoleGroupConfig {
+  traderGroupName: string
+  chaGroupName: string
+}
```
mushrafmim, line 8: Can't we make this an array instead? If we are to introduce a new group, then we need to update the codes in a lot of places? violating Open Closed Principle.

start-dev.sh
```diff
@@ -104,9 +104,11 @@ TRADER_IDP_CLIENT_ID="${TRADER_IDP_CLIENT_ID:-TRADER_PORTAL_APP}"
 NPQS_IDP_CLIENT_ID="${NPQS_IDP_CLIENT_ID:-OGA_PORTAL_APP_NPQS}"
 FCAU_IDP_CLIENT_ID="${FCAU_IDP_CLIENT_ID:-OGA_PORTAL_APP_FCAU}"
 IRD_IDP_CLIENT_ID="${IRD_IDP_CLIENT_ID:-OGA_PORTAL_APP_IRD}"
-IDP_SCOPES="${IDP_SCOPES:-openid,profile,email}"
+IDP_SCOPES="${IDP_SCOPES:-openid,profile,email,group,role}"
 IDP_PLATFORM="${IDP_PLATFORM:-AsgardeoV2}"
 SHOW_AUTOFILL_BUTTON="${SHOW_AUTOFILL_BUTTON:-true}"
+TRADER_IDP_TRADER_GROUP_NAME="${TRADER_IDP_TRADER_GROUP_NAME:-Traders}"
+TRADER_IDP_CHA_GROUP_NAME="${TRADER_IDP_CHA_GROUP_NAME:-CHA}"
```
mushrafmim, line 111: Can't we make this an array like `IDP_SCOPES`?

portals/apps/trader-app/src/screens/UnauthorizedScreen.tsx
```diff
@@ -0,0 +1,36 @@
...
+                    </p>
+                    <div className="mt-8 flex items-center justify-center">
+                        <SignedIn>
+                            <Button onClick={handleSignOut} size="3" style={{ cursor: 'pointer' }}>
+                                Sign out
+                            </Button>
+                        </SignedIn>
+                        <SignedOut>
+                            <SignInButton />
+                        </SignedOut>
```
mushrafmim: Is there any chance this code to be evaluated? Since we are displaying this screen only when a person is `unauthorized`.

mushrafmim (changes requested):

portals/apps/trader-app/src/screens/UnauthorizedScreen.tsx
sthanikan2000: Don't we need the top bar?
What is best in the UI perspective?
Only showing the Unauthorized screen with signout enough?

start-dev.sh
sthanikan2000: These are expected claims which should match exactly.

portals/apps/trader-app/src/runtimeConfig.ts
sthanikan2000: If we introduce a group. We should define the access permissions anyway. So will update in many places

portals/apps/trader-app/src/screens/UnauthorizedScreen.tsx
sthanikan2000: yeah can be removed

sthanikan2000:
@mushrafmim Basically what implemented here is, we will first check whether the Groups decoded from the ID match any of the Expected Group claims name. Then if no match then unauthorized screen, else we map the Expected Groups name given in the environment configuration with the Constants defined in the frontend role for RBAC. These constants you can define as lower-case. It is not mandatory to define it as lowercase in the IdP

portals/apps/trader-app/src/screens/UnauthorizedScreen.tsx
mushrafmim: No, I mean, instead of introducing another TopBar called `UserOnlyTopBar`, Can't you place the same `TopBar` and hide the non related stuff inside `TopBar` (Conditionally render only the required components). So, there's no 2 TopBars. We are using the same component type.

sthanikan2000: okay I'll do that

portals/apps/trader-app/src/runtimeConfig.ts
sthanikan2000: I'll check on the best practice to do this

portals/apps/trader-app/src/screens/UnauthorizedScreen.tsx
sthanikan2000: I updated the Unauthorized screen as simple

I don't want to waste time on UI. Let's fix UI later

I remove the TopBar and removed the code for User only Top bar

portals/apps/trader-app/src/screens/UnauthorizedScreen.tsx
sthanikan2000: removed

portals/apps/trader-app/src/runtimeConfig.ts
sthanikan2000: If I keep this lowercase, If someone don't specify the correct name of the groups in the env config, they will not able to get in to the system. What should I do here?

sthanikan2000:
@mushrafmim Can you go through the `portals/apps/trader-app/src/utils/roleMapper.ts` and `portals/apps/trader-app/src/runtimeConfig.ts`? According to the logic which I implemented here. We need them as separate string instead of Array. What do you think?

https://github.com/sthanikan2000/nsw/blob/63c482b31ce143f8c43edc727c450b232267a28f/portals/apps/trader-app/src/utils/roleMapper.ts

mushrafmim (approved):
LGTM

Aravinda-HWK (approved):

PR #318, opened 2026-04-05
feat(template): implement template registry with form handler and validation

sthanikan2000:
## Summary
This PR introduces only the initial OGA template module foundation. It adds the base abstractions and validation needed for a plugin-based template system, without changing current service wiring yet.

## Type of Change
- [x] New feature (foundation layer)

## Changes Made
- Added template package models: `model.go`
- Added common template handler interface: `interface.go`
- Added FORM handler with validation logic: `form.go`
- Implemented TemplateRegistry for loading, validation, lookup, and default template resolution: `registry.go`
- Added focused unit tests for template registry behavior: `registry_test.go`
- Added integration test with realistic mock template data showing end-to-end usage of the template package: `integration_test.go`

## Testing
- [x] All existing tests pass

## Additional Notes
This PR is intentionally scoped to the newly created template module only.

Follow-up PRs will cover:
- Replace formStore with templateRegistry
- Migration of template files
- Implement MarkdownHandler
- Refactor OGA Portal
- Update docs

### DIFF
--- a/oga/internal/template/interface.go
+++ b/oga/internal/template/interface.go
@@ -1,0 +1,12 @@
+package template
+
+import (
+	"context"
+	"encoding/json"
+)
+
+type TemplateHandler interface {
+	Type() TemplateType
+	Validate(content map[string]any) error
+	Process(ctx context.Context, content json.RawMessage, data map[string]any) (any, error)
+}
--- a/oga/internal/template/form.go
+++ b/oga/internal/template/form.go
@@ -1,0 +1,45 @@
+package template
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+type FormHandler struct{}
+
+func NewFormHandler() *FormHandler {
+	return &FormHandler{}
+}
+
+func (h *FormHandler) Type() TemplateType {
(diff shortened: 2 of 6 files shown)


ginaxu1 (approved):
Lgtm

oga/internal/template/interface.go
```diff
@@ -0,0 +1,12 @@
+package template
+
+import (
+	"context"
+	"encoding/json"
+)
+
+type TemplateHandler interface {
+	Type() TemplateType
+	Validate(content map[string]any) error
```
mushrafmim, line 10: `Validate` is to perform the validation on the schema while registering a schema.

Shouldn't we have another function to perform validation on data? Let's say we have a markdown template handler. Shouldn't we have another function, that validates the data arriving to be processed, whether it aligns with the data format?

Or For the FormTemplateHandler to perform validation over whether it aligns with the form's schema?

mushrafmim (changes requested):

oga/internal/template/interface.go
sthanikan2000: Yeah we should have that function. Will update this PR

PR #285, opened 2026-03-25
284 helm updates

ginaxu1:
Closes #284. Reference [deployments/helm/DEPLOYMENT.md](https://github.com/ginaxu1/nsw/blob/284-helm-updates/deployments/helm/DEPLOYMENT.md)

## Summary
This PR provides the deployment configuration and container security hardening required to run the NSW platform on OpenShift (Akaza GovCloud). Changes address strict resource quotas, Security Context Constraints (SCCs), and consolidate Helm chart management under `deployments/helm/` for deployment across dev, staging, and prod

Portal Routes 
- [dev-trader-app-national-single-window-platform.apps.sovecloud1.akaza.lk](https://dev-trader-app-national-single-window-platform.apps.sovecloud1.akaza.lk/login)   
- [dev-oga-fcau-national-single-window-platform.apps.sovecloud1.akaza.lk](https://dev-oga-fcau-national-single-window-platform.apps.sovecloud1.akaza.lk/login)                    
- [dev-oga-npqs-national-single-window-platform.apps.sovecloud1.akaza.lk](https://dev-oga-npqs-national-single-window-platform.apps.sovecloud1.akaza.lk/login)
- [dev-oga-ird-national-single-window-platform.apps.sovecloud1.akaza.lk](https://dev-oga-ird-national-single-window-platform.apps.sovecloud1.akaza.lk/login)

Backend / Other Routes
- dev-oga-fcau-backend-national-single-window-platform.apps.sovecloud1.akaza.lk
- dev-oga-npqs-backend-national-single-window-platform.apps.sovecloud1.akaza.lk
- dev-oga-ird-backend-national-single-window-platform.apps.sovecloud1.akaza.lk
- dev-api-national-single-window-platform.apps.sovecloud1.akaza.lk
- temporal-ui-national-single-window-platform.apps.sovecloud1.akaza.lk
- [idp-thunder-national-single-window-platform.apps.sovecloud1.akaza.lk](https://idp-thunder-national-single-window-platform.apps.sovecloud1.akaza.lk/console/)
#### 1. Environment-Specific Helm Values
* Files: `*/values.yaml`, `*/values-dev.yaml`, `*/values-staging.yaml`, `*/values-prod.yaml`
* Details: Abstracted configurations into base `values.yaml` files and environment-specific overrides. Variables like external URLs, IDP issuers, and database coordinates are now dynamically injected based on the target environment, keeping our deployment logic completely DRY.

#### 2. K8s Secret Dependencies & Image Pulling
[...]

PR #345, opened 2026-04-10
feat(Helm): add Charts for NSW backend and database

ginaxu1:
## Summary
Already have all services running. 
Breaking up #285 into smaller pull requests, this one includes Helm Chart for NSW backend (dev-nsw-api-national-single-window-platform.apps.sovecloud1.akaza.lk ) and database. These charts are designed to support a "zero-touch" GitOps workflow by sourcing base templates from this repository and environment-specific overrides from the `nsw-gitops` repository.

## Database Migration Workflow
To add a new migration script and have Helm execute it automatically:

1. Create Migration Scripts
Add sequential `.up.sql` and `.down.sql` files to:
`backend/internal/database/migrations/` (e.g., `013_add_feature.up.sql`)

2. Rebuild & Push Migration Image
```bash
docker build -t ghcr.io/opennsw/nsw-api-migrations:latest -f backend/migrations.Dockerfile .
docker push ghcr.io/opennsw/nsw-api-migrations:latest

### CODE REVIEW

File: deployments/helm/nsw-api/values.yaml [hunk 1]
```diff
@@ -26,22 +41,28 @@ config:
...
+  # servicePrefix is the Helm release name prefix used to construct internal OGA backend service URLs.
+  # It MUST match the release name used when deploying the OGA backend charts.
+  # Example: release "dev-nsw-api" → servicePrefix "dev" → resolves to "dev-oga-fcau-backend"
+  servicePrefix: "dev"
 
   cors:
-    allowedOrigins:
-      - "https://trader-portal-national-single-window-platform.lgc.gov.lk"
-      - "https://npqs-portal-national-single-window-platform.lgc.gov.lk"
+    allowedOrigins: "https://trader-app-national-single-window-platform.apps.sovecloud1.akaza.lk,https://oga-fcau-national-single-window-platform.apps.sovecloud1.akaza.lk,https://oga-ird-national-single-window-platform.apps.sovecloud1.akaza.lk,https://oga-npqs-national-single-window-platform.apps.sovecloud1.akaza.lk"
```
sthanikan2000:
```suggestion
    allowedOrigins: "https://trader-app-national-single-window-platform.apps.sovecloud1.akaza.lk"
```

deployments/helm/nsw-api/values.yaml
```diff
@@ -26,22 +41,28 @@ config:
...
+    jwksUrl: "http://idp-thunder-service.national-single-window-platform.svc.cluster.local:8090/oauth2/jwks"
+    issuer: "https://idp-thunder-national-single-window-platform.apps.sovecloud1.akaza.lk"
     audience: "TRADER_PORTAL_APP"
     clientId: "TRADER_PORTAL_APP"
+    # Set to true if the IDP is using a self-signed cert internally (common in GovCloud)
     insecureSkipTlsVerify: false
 
-  serviceUrl: "https://nsw-api-national-single-window-platform.lgc.gov.lk"
+  serviceUrl: "https://nsw-api-national-single-window-platform.apps.sovecloud1.akaza.lk/api/v1"
+  internalServiceUrl: "http://nsw-api.national-single-window-platform.svc.cluster.local"
```
sthanikan2000: What is the reason for this `internalServiceUrl`?

deployments/helm/nsw-api/values.yaml
sthanikan2000, line 1: Check with the values in `values-staging.yaml` and `values-dev.yaml` as well

deployments/helm/nsw-api/templates/deployment.yaml
```diff
@@ -44,34 +52,49 @@ spec:
...
+            {{- if .Values.config.auth.credentialsSecretName }}
+            - name: AUTH_CLIENT_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.config.auth.credentialsSecretName | quote }}
+                  key: "clientSecret"
+            {{- else }}
+            - name: AUTH_CLIENT_SECRET
+              value: {{ .Values.config.auth.clientSecret | quote }}
+            {{- end }}
```
sthanikan2000: What is the reason for the setting CLIENT_SECRET? We'll not use client_secret in the JWT token validation

sthanikan2000 (changes requested):

sthanikan2000:
@ginaxu1 You're missing an environment variable USE_WORKFLOW_MANAGER_V2 

Check with the `backend/internal/config/config.go` and check whether you specified all the environment variables

deployments/helm/nsw-api/values.yaml
ginaxu1: `internalServiceUrl` typically for communication within the cluster. in this case `oga-backend` uses this URL to call NSW backend for
1. download URL `GET /api/v1/uploads/{key}` - uses the internal cluster URL to keep traffic private and fast
2. submitting review decisions: oga-backend sends POST request to callback `/api/v1/tasks`

deployments/helm/nsw-api/templates/deployment.yaml
ginaxu1: good catch, removed. NSW backend validates incoming JWT tokens using the `AUTH_JWKS_URL`. this process only needs Issuer and Audience, no need for client secret

deployments/helm/nsw-api/values.yaml
ginaxu1: Need this for now because 1. file upload not yet using S3 and 2. OGA needs to call NSW backend for file uploads; if we make this change to only allow trader-app, then OGA portals will have CORS blocking (won't be able to upload or view attached docs)

https://github.com/OpenNSW/nsw/pull/337 decouples OGA from using NSW backend upload handler. once that is merged and S3 set up in prod, can remove the OGA urls from above

deployments/helm/nsw-api/values.yaml
sthanikan2000: but ideally OGAs will hit NSW through the Public endpoints. There will be no internalServiceUrl we need I guess.
Even though you specify the internal Urls. still the other Urls can be used right?

deployments/helm/nsw-api/values.yaml
sthanikan2000: **Keeping the OGA origins in the CORS make sense and mandatory if we're using `local_fs`.**

But I didn't get what you're telling about the #337.
If we're using `local_fs` driver, we need to add all the OGA origins to the CORS. If we're only using the S3 only we don't need to specify it in CORS.
I think we need multiple CORS middleware configured or ALLOWED_ORGINS configurable. That should be done as separate PR. Let's discuss about the implementation details offline.

.github/workflows/release.yml
```diff
@@ -12,8 +13,13 @@ name: Release - Build and Publish All Images
 
 on:
   push:
+    branches:
+      - main
     tags:
       - 'v*.*.*'  # matches semantic version tags: v1.0.0, v1.2.3-rc.1, etc.
+  pull_request:
+    branches:
+      - main
```
sthanikan2000:
```suggestion
tags:
- 'v*.*.*'  # matches semantic version tags: v1.0.0, v1.2.3-rc.1, etc.
```
Revert this. We're not going to release per each push to main.

sthanikan2000:
@ginaxu1 This PR seems not up to date. Can you take a look? also address all the comments

sthanikan2000 (changes requested):
We don't need nsw-db helm charts

Aravinda-HWK (approved):
LGTM.

sthanikan2000 (approved):

PR #350, opened 2026-04-11
feat(Helm): add IDP Chart + Dockerfile

ginaxu1:
## Summary
Already have all services running. 
Breaking up #285 into smaller pull requests, this one includes Helm Chart for Thunder [https://idp-thunder-national-single-window-platform.apps.sovecloud1.akaza.lk/console](https://idp-thunder-national-single-window-platform.apps.sovecloud1.akaza.lk/console)

## Changes
Add umbrella chart that pulls Thunder IDP as a dependency:
deployments/helm/idp/Chart.lock
deployments/helm/idp/Chart.yaml
deployments/helm/idp/values-dev.yaml
deployments/helm/idp/values-staging.yaml

```bash
# Verify the files
ls -F deployments/helm/idp/
```


deployments/helm/idp/values-staging.yaml
```diff
@@ -0,0 +1,37 @@
+# Staging Overrides for idp-thunder umbrella
+thunder:
+  configuration:
+    server:
+      publicUrl: "https://idp-thunder-national-single-window-platform.apps.sovecloud1.akaza.lk"
```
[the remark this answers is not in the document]
ginaxu1: updating to shared values.yaml, dev and staging sharing one instance of Thunder so no need for -dev and -staging.yaml

deployments/docker/idp-custom/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ghcr.io/asgardeo/thunder:0.29.0
```
Aravinda-HWK: @ginaxu1 We are using 0.35.0 currently in the system.

deployments/docker/idp-custom/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# Switch to root temporarily to modify permissions
+USER root
+
+# Replace the broken binary
+COPY ./fixed-binary /opt/thunder/bin/binary
```
Aravinda-HWK: Could you please explain to me the reason for copy local binary file into the existing binary file?
I would like to know what this fixed-binary file means

Aravinda-HWK (changes requested):

deployments/docker/idp-custom/Dockerfile
ginaxu1: removed this file; was originally in an older commit to work around an issue in the base image `ghcr.io/asgardeo/thunder:0.29.0`

deployments/helm/idp/files/02-sample-resources.sh
Aravinda-HWK, line 1: I checked the diff between this file and the existing [sample-resources](https://github.com/OpenNSW/nsw/blob/main/idp/02-sample-resources.sh) file IDP folder. Can you check that, and I am not sure whether we need to duplicate the same things in two different places.

sthanikan2000 (changes requested):
We don't need to define helm for the Thunder. We need to bring the Thunder with default setup provided by Thunder team. After that we need to run an initial setup script to bring the required resources.
This is not the blocker for the Automation setup. Because we're not going to update thunder version every time.  So Let's skip the auto setup for the infra for now.

idp/common.sh
Aravinda-HWK, line 1: Do we need `common.sh` because I think this is coming with the Thunder, and we don't need to update anything here.

Aravinda-HWK (changes requested):

ginaxu1:
@sthanikan2000  agree, we will skip auto-GitOps sync for the `nsw-infra-*` deployments (which has Temporal, Thunder, nsw-db) . Best to only redeploy Thunder if new official Version updated

Actually the Official Thunder default setup fails on our clusters for these reasons:
1. SCC Violations: upstream setup uses hardcoded UIDs (e.g., 10001) which are blocked by the `restricted-v2` Security Context Constraint. this chart is patched to allow OpenShift to assign random UIDs
2. Port mismatches: upstream targets port `8090`, but the service logic and probes must align with the actual listener on `8080` to avoid 503 errors
3. SubPath mounting: required for `signing.key` and `crypto.key` secrets to avoid mounting conflicts with existing file structures


So best to keep `deployments/helm/idp` as starting point for those occasional Thunder re-deployments to `nsw-infra-*` (dev/staging/prod)

idp/common.sh
ginaxu1: actually `02-sample-resources.sh` looks for common.sh in its own directory to load the thunder_api_call functions. So the `idp/common.sh` needed

deployments/helm/idp/files/02-sample-resources.sh
ginaxu1: so this is because Helm can only look into chart's directory and cannot reach files in the root idp/ folder
and this copy needed for automated [post-install job](https://github.com/OpenNSW/nsw/pull/350/changes#diff-2909d9e45b0845272853a015dd864b9b8ff4054ec6e00b4f7a52e959ff83e438)

sthanikan2000 (changes requested):
First revert the changes you did in the `idp/**`, `deployments/helm/idp/charts/**`, and ``deployments/helm/idp/files/**`. These are unnecessary to bring default Thunder instance. I'll take a look at the other changes.

PR #361, opened 2026-04-16
feat(nsw-core,auth): support mixed SPA and M2M principals with multi-client auth config

### LINKED ISSUE #343: [NSWCore][Security] Auth foundation for mixed SPA and M2M auth
## Problem
The backend auth layer currently behaves like a single-principal JWT system. It does not yet carry enough identity data or config structure to safely onboard multiple trusted external systems by configuration.

## Proposed Solution
Extend the auth foundation so the backend can reliably distinguish SPA and M2M principals and validate trusted inbound callers by config.

Scope:
- Extend auth claims/context to capture principal identity needed for authorization decisions.
- Support trusted inbound service configuration keyed by `client_id` and the required matching token identity fields.
- Preserve existing SPA auth behavior.
- Keep authentication separate from authorization.

## Acceptance Criteria
- The auth layer can represent SPA and M2M principals without ambiguity.
- Trusted external callers can be added by config, not code changes.
- Invalid or incomplete trusted-service config fails fast at startup.
- Existing SPA token behavior remains unchanged.

## Alternatives
- Hard-code allowlists in middleware. This does not scale and is not config-driven.
- Make separate auth systems for SPA and M2M. This increases duplication and maintenance cost.

## Relationships
- `parent_issue`: #12

### LINKED ISSUE #365: [NSWCore][Authn] Persist user context on first auth
### Problem

When a user authenticates successfully but no matching user context exists in the database, the middleware currently logs the miss and continues with `UserContext = nil`.

That is safe for request handling, but it leaves a gap in the auth flow: new users are not automatically persisted, so the same lookup miss can happen repeatedly until something else creates the context out of band.

### Proposed Solution

Create the user context the first time a valid user principal is seen and no record exists yet.

Example implementation idea:

```go
if errors.Is(err, gorm.ErrRecordNotFound) {
	slog.Debug("no stored user context, creating one for first-time user",
		"user_id", principal.UserPrincipal.UserID)

	newUserCtx := &UserContext{
		UserID: principal.UserPrincipal.UserID,
		UserContext: json.RawMessage(`{
			"email": "` + principal.UserPrincipal.Email + `",
			"ouHandle": "` + principal.UserPrincipal.OUHandle + `",
			"ouId": "` + principal.UserPrincipal.OUID + `"
		}`),
	}

	if err := authService.CreateUserContext(newUserCtx); err != nil {
		slog.Error("failed to create user context for new user",
			"user_id", principal.UserPrincipal.UserID,
			"error", err,
		)
		w.Header().Set("Content-Type", "application/json")
		w.WriteHeader(http.StatusInternalServerError)
		_, _ = w.Write([]byte(`{"error":"internal_server_error","message":"failed to create user context"}`))
		return
	}
}
```

### Alternatives

- Keep the current behavior and rely on an external bootstrap job or admin workflow to create user contexts.
- Create user context lazily in a separate onboarding flow instead of in auth middleware.
- Defer context creation entirely and treat the first authenticated request as read-only until a dedicated provisioning step runs.

sthanikan2000:
## Summary

This PR introduces the authentication foundation for mixed SPA and M2M traffic in NSW.

It also provides the backend auth groundwork required by PR #353 (`feat(oauth2): implement NSW outbound API integration with OAuth2 support`).

It now also includes first-auth user-context persistence, so a valid user principal creates its database context the first time it is seen.

## Type of Change

- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement

## Changes Made

- Updated auth configuration from single-client to multi-client support:
  - `AUTH_CLIENT_ID` -> `AUTH_CLIENT_IDS`
- `AUTH_AUDIENCE` standardized to `NSW_API`
- Enhanced token parsing to support grant-aware principal extraction:
  - `authorization_code` -> user principal
  - `client_credentials` -> client principal
- Added auth context support for both user and client identities.
- Kept authentication concerns separate from authorization decisions.
- Updated workflow routers (consignment and pre-consignment handlers) to safely dereference authenticated `UserID` for SPA flows.
- Persisted user context on first authenticated user request when no record exists yet.
- Added/updated tests for grant type parsing and principal behavior.

## Testing

- [x] All existing tests pass

## Checklist

## Related Issues

Closes #343
Closes #365
- The auth layer now represents SPA and M2M principals without ambiguity.
- Trusted callers are managed via configuration (`AUTH_CLIENT_IDS`) rather than code changes.
- Invalid or incomplete auth configuration fails fast during startup validation.
- Existing SPA behavior remains intact while enabling M2M expansion.
- First-time authenticated users now get a persisted user context instead of repeatedly hitting the miss path.

### Breaking Change Notice

This PR is a breaking change for environments still using legacy auth variables.

### Required Configuration Changes

- Replace `AUTH_CLIENT_ID` with `AUTH_CLIENT_IDS` (comma-separated list).
- Ensure `AUTH_AUDIENCE=NSW_API` is set in all runtime environments.
- Update any deployment manifests, helm values, and secrets that still reference the previous auth variable names.

## Deployment Notes

- Deployment configuration must be updated before rollout.
- Existing deployments that do not provide `AUTH_CLIENT_IDS` will fail startup validation.
- Coordinate backend and environment variable updates together to avoid startup/auth failures.


backend/internal/auth/token_parser.go
```diff
@@ -137,7 +159,7 @@ func (te *TokenExtractor) ExtractClaimsFromHeader(authHeader string) (*Extracted
 	parsedToken, err := jwt.ParseWithClaims(tokenString, claims, te.keyFunc,
 		jwt.WithValidMethods([]string{"RS256", "RS384", "RS512"}),
 		jwt.WithIssuer(te.issuer),
-		jwt.WithAudience(te.audience),
+		// jwt.WithAudience(te.audience), // TODO: Once Thunder supports defining audience claim, add this validation back.
```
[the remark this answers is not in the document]
sthanikan2000: This was disabled intentionally. #364 created to address this in future.

backend/internal/auth/service.go
```diff
@@ -62,11 +57,6 @@ func (as *AuthService) GetUserContext(userID string) (*UserContext, error) {
 //
 //	newContext := json.RawMessage(`{"company": "Acme Inc", "role": "exporter"}`)
 //	err := authService.UpdateUserContext("TRADER-001", newContext)
-//
```
ginaxu1, line 65: notice you deleted TODO; double check you implemented the TODOs in `backend/internal/auth/service.go`

ginaxu1 (changes requested):

backend/internal/workflow/router/consignment_router.go
ginaxu1, line 126: missing return here

ginaxu1, line 55: missing return here

ginaxu1 (changes requested):

backend/internal/auth/middleware.go
```diff
@@ -51,42 +46,87 @@ func Middleware(authService *AuthService, tokenExtractor *TokenExtractor) func(h
...
+						// 	OUID:     principal.UserPrincipal.OUID,
+						// }
+						// err = authService.CreateUserContext(newUserCtx)
+						// if err != nil {
+						// 	slog.Error("failed to create user context for new user",
+						// 		"user_id", principal.UserPrincipal.UserID, "error", err)
+						// } else {
+						// 	slog.Info("created new user context for user",
+						// 		"user_id", principal.UserPrincipal.UserID)
+						// }
```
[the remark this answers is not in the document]
sthanikan2000: Created issue #365.
Removed the code block and kept only the TODO comment

backend/internal/auth/service.go
sthanikan2000: Good catch. I though each comments related to JWT should be deleted. But there are something needs to be modified such that, JWT fully implemented and there are enhancements need to be done.
I'll update the comments accordingly.

backend/internal/workflow/router/consignment_router.go
sthanikan2000: This is the last executable code block. Are we need to put `return` statements explicitly? What is the best practice?

ginaxu1: it seems omitting return after http.Error is a Go anti-pattern. can lead to bugs if code is ever appended to the function

ginaxu1, line 141: should you also add check for
```
if authCtx.UserID == nil {
    return http.Error(w, "M2M tokens cannot initialize consignments", 403) 
}
```

ginaxu1 (changes requested):

backend/internal/auth/token_parser.go
```diff
@@ -137,7 +159,7 @@ func (te *TokenExtractor) ExtractClaimsFromHeader(authHeader string) (*Extracted
 	parsedToken, err := jwt.ParseWithClaims(tokenString, claims, te.keyFunc,
 		jwt.WithValidMethods([]string{"RS256", "RS384", "RS512"}),
 		jwt.WithIssuer(te.issuer),
-		jwt.WithAudience(te.audience),
+		// jwt.WithAudience(te.audience), // TODO: Once Thunder(IdP) supports defining audience claim, add this validation back.
```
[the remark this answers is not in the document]
sthanikan2000: https://github.com/OpenNSW/nsw/pull/361#discussion_r3097766335

backend/internal/workflow/router/consignment_router.go
sthanikan2000: These will be handle by the `Authorization` layer will be implemented after finishing the `Authentication` layer.

ginaxu1 (dismissed):

backend/internal/workflow/router/consignment_router.go
```diff
@@ -91,26 +100,42 @@ func (c *ConsignmentRouter) HandleGetConsignments(w http.ResponseWriter, r *http
...
+			http.Error(w, "failed to resolve default CHA profile", http.StatusInternalServerError)
 			return
 		}
 		filter.ChaID = &cha.ID
 	case "trader":
-		filter.TraderID = &authCtx.UserID
+		if authCtx.UserID == nil {
+			http.Error(w, "User ID is required in auth context", http.StatusBadRequest)
+			return
+		}
```
mushrafmim: Why email validation should be added here? Shouldn't it be inside the token parsing logic? and this code is inconsistent, while the context has the direct access to `UserID`, it doesn't have the access to the `ChaID`, either both should be in the token or nothing?

And why can't we obtain the Cha from the `UserID`?

backend/internal/auth/context.go
```diff
@@ -16,12 +16,16 @@ func (t *UserContext) TableName() string {
...
-	UserID      string       `json:"userId"`
-	Email       string       `json:"email"`
-	OUHandle    string       `json:"ouHandle"`
+	UserID      *string      `json:"userId,omitempty"`
+	Email       *string      `json:"email,omitempty"`
+	OUHandle    *string      `json:"ouHandle,omitempty"`
+	OUID        *string      `json:"ouId,omitempty"`
+	ClientID    *string      `json:"clientId,omitempty"`
 	UserContext *UserContext `json:"userContext,omitempty"`
 }
```
mushrafmim, line 33: The current structure is confusing, we have AuthContext.UserID and then AuthContext.UserContext.UserID. I would suggest a pattern like this which is cleaner and easy to use.
```go
type PrincipalType string

  const (
      UserPrincipalType   PrincipalType = "user"
      ClientPrincipalType PrincipalType = "client"
  )

  type AuthContext struct {
      Type   PrincipalType
      User   *AuthenticatedUser
      Client *AuthenticatedClient
  }

  type AuthenticatedUser struct {
      ID        string
      Email     string
      OUHandle  string
      OUID      string
      Context   map[string]any
  }

  type AuthenticatedClient struct {
      ID string
  }
```

mushrafmim (changes requested):

backend/internal/auth/context.go
sthanikan2000: Yeah this make sense. The previous one was confusing. I bring the multi-client support and fixed the places where the error thrown. But AuthContext is also part of the `authn` pkg. I'lll redo it considering your suggestion

sthanikan2000: @mushrafmim shall I rename the `auth` pkg as `authn`?

backend/internal/workflow/router/consignment_router.go
sthanikan2000: There are many inconsistency in the `router`. I'll not touch everything in this PR. But will update the codes related to the `AuthContext`

backend/internal/auth/context.go
mushrafmim: Ok.

backend/internal/workflow/router/consignment_router.go
mushrafmim: Sure, Need to perform a proper cleanup, let's consider it out of scope of this project. But I think you can remove the email validation, because email validation is performed in the auth middleware.

sthanikan2000: Yeah I removed that. Will let you know once the PR is ready to review again

backend/internal/auth/context.go
sthanikan2000: I'll create an issue for that renaming. If I do that changes in this PR, it will change a lot

ginaxu1 (dismissed):

backend/internal/auth/context.go
sthanikan2000: created #369

backend/internal/workflow/router/consignment_router.go
sthanikan2000: @mushrafmim PR is ready for review

backend/internal/auth/context.go
```diff
@@ -7,32 +7,39 @@ import (
 
 // UserContext represents a user's stored context in the database.
 type UserContext struct {
-	UserID      string          `gorm:"type:varchar(100);column:user_id;primaryKey;not null" json:"userId"`
-	UserContext json.RawMessage `gorm:"type:jsonb;column:user_context;serializer:json;not null" json:"userContext"`
+	UserID   string          `gorm:"type:varchar(100);column:user_id;primaryKey;not null" json:"userId"`
+	Email    string          `gorm:"type:varchar(255);column:email" json:"email"`
+	OUHandle string          `gorm:"type:varchar(255);column:ou_handle" json:"ouHandle"`
+	OUID     string          `gorm:"type:varchar(255);column:ou_id" json:"ouId"`
+	NSWData  json.RawMessage `gorm:"type:jsonb;column:nsw_data" json:"nswData"`
```
mushrafmim, line 14: Why have we added these `gorm` stuff? Is there any particular reason?

sthanikan2000: We're storing the UserAuthContext in the database. This was the previous functionality to store any NSW specific data related to a Trader. If we remove this we need to rewrite the initialize consignment logic

sthanikan2000:
@ginaxu1 @mushrafmim I fixed some issues related to the test cases https://github.com/OpenNSW/nsw/pull/361/changes/52a601d8fc89f4ac5f1ceea8cf4e2b9ee4844451

This came because of we merged the presigned URL PR #362

backend/internal/auth/context.go
mushrafmim: I think User level data might need extra stuff as well. Like User's mobile number (and so on). In that case, I think we should decouple the database definition from the Context (of Auth). Let's consider this in the subsequent PRs.

mushrafmim (approved):

ginaxu1 (approved):

PR #362, opened 2026-04-16
feat(upload part 1): implement pre-signed URLs in upload service 

ginaxu1:
## Summary
Implement backend changes for migrating the file upload mechanism to use pre-signed URLs. This is Phase 1 of 3 part rollout to transition the platform gracefully. Upload service now securely generates and returns pre-signed upload URLs while retaining full backwards compatibility with the existing legacy `multipart/form-data` upload endpoints for seamless transition

**Note `PUT local-dev` endpoint's exclusive to local development**
-  In prod, will be 1 API call to POST /uploads. Then, frontend uploads to the URL returned by backend (via `upload_url`). File is NOT ever sent to backend.
- in Local Dev environment, frontend _does_ send the file to backend (via the `PUT local-dev` endpoint). But this is a necessary evil if we don't run tools like seaweedfs / MinIO locally! `PUT local-dev` handler mimic S3's behavior so that we can test the exact same frontend logic

## Changes
- New API Endpoint Flow: add handling for JSON-based requests to `/uploads` to generate pre-signed upload URLs with strong time-to-live (TTL) and metadata constraints
- HMAC Signatures: implement HMAC-SHA256 tokens using null byte (`\x00`) separators in `token.go` to sign security constraints (size limit, expiration, content type) and mitigate payload collision risks
- Local Dev Mocking: extend `LocalFSDriver` to reliably mimic S3 pre-signed URL behaviors for local development without actual object storage
- Security Hardening: enforce a `1MB` max request body limit (`http.MaxBytesReader`) on the JSON metadata endpoint to prevent DoS attacks
- Error Handling: improve error checking by using `errors.As` to properly map `http.MaxBytesError` instances to HTTP 413 (Payload Too Large) responses

## Testing
- Verified backwards compatibility with legacy `multipart/form-data` clients
- Uploaded valid and invalid file sizes, ensuring the `1MB` metadata limit blocks large payloads with an HTTP 413 error

### DIFF
--- a/api/openapi-uploads.yaml
+++ b/api/openapi-uploads.yaml
@@ -29,19 +29,24 @@ paths:
       requestBody:
         required: true
         content:
-          multipart/form-data:
+          application/json:
             schema:
               type: object
               required:
-                - file
+                - filename
+                - mime_type
+                - size
               properties:
-                file:
+                filename:
                   type: string
-                  format: binary
-                  description: The file to upload
-            encoding:
-              file:
-                contentType: application/octet-stream
+                  description: The original filename
+                mime_type:
+                  type: string
+                  description: The MIME type of the file
+                size:
+                  type: integer
+                  format: int64
+                  description: The file size in bytes
       responses:
         "200":
           description: File uploaded successfully
@@ -53,9 +58,10 @@ paths:
                 id: "550e8400-e29b-41d4-a716-446655440000"
                 name: "document.pdf"
                 key: "550e8400-e29b-41d4-a716-446655440000.pdf"
-                url: "/bucket/550e8400-e29b-41d4-a716-446655440000.pdf"
+                url: "/api/v1/uploads/550e8400-e29b-41d4-a716-446655440000.pdf/content"
+                upload_url: "/api/v1/uploads/local-put/550e8400-e29b-41d4-a716-446655440000.pdf?token=..."
                 size: 1024000
-                mimeType: "application/pdf"
+                mime_type: "application/pdf"
         "400":
           description: Bad request - file is required or invalid
           content:
@@ -75,11 +81,10 @@ paths:
 
   /uploads/{key}:
     get:
-      summary: Download File
+      summary: Get File Download URL
       description: >
-        Download a previously uploaded file by its unique key.
-        Returns the file content with the appropriate content type.
-      operationId: downloadFile
+        Fetches a presigned or limited-time URL to securely download the file content.
+      operationId: getDownloadUrl
       tags:
         - Uploads
       parameters:
@@ -92,20 +97,20 @@ paths:
           example: "550e8400-e29b-41d4-a716-446655440000.pdf"
       responses:
         "200":
-          description: File retrieved successfully
+          description: Download URL generated successfully
           content:
-            application/octet-stream:
+            application/json:
               schema:
-                type: string
-                format: binary
-            application/pdf:
-              schema:
-                type: string
-                format: binary
-            image/*:
-              schema:
-                type: string
-                format: binary
+                type: object
+                properties:
+                  download_url:
+                    type: string
+                    description: The presigned URL to download the actual file contents
+                    example: "/api/v1/uploads/550e8400-e29b-41d4-a716-446655440000.pdf/content?token=12345&expiresAt=123"
+                  expires_at:
+                    type: integer
+                    description: Unix timestamp indicating when the link expires
+                    example: 1713256000
         "400":
           description: Bad request - key is required or invalid
           content:
@@ -168,11 +173,9 @@ components:
       type: object
       required:
         - id
-        - name
         - key
-        - url
         - size
-        - mimeType
+        - mime_type
       properties:
         id:
           type: string
@@ -190,13 +193,17 @@ components:
         url:
           type: string
           description: Public URL to access the file
-          example: "/bucket/550e8400-e29b-41d4-a716-446655440000.pdf"
+          example: "/api/v1/uploads/550e8400-e29b-41d4-a716-446655440000.pdf/content"
+        upload_url:
+          type: string
+          description: Presigned or local URL for uploading the file content
+          example: "/api/v1/uploads/local-put/550e8400-e29b-41d4-a716-446655440000.pdf?token=..."
         size:
           type: integer
           format: int64
           description: File size in bytes
           example: 1024000
-        mimeType:
+        mime_type:
           type: string
           description: MIME type of the uploaded file
           example: "application/pdf"
--- a/api/openapi-uploads.yaml
+++ b/api/openapi-uploads.yaml
@@ -173,6 +173,7 @@ components:
       type: object
       required:
         - id
+        - name
         - key
         - size
         - mime_type
--- a/api/openapi-uploads.yaml
+++ b/api/openapi-uploads.yaml
@@ -30,6 +30,9 @@ paths:
         required: true
         content:
           application/json:
+            description: >
+              New pre-signed URL flow. Returns upload metadata including a pre-signed
+              URL to PUT the file contents directly to storage.
             schema:
               type: object
               required:
@@ -47,6 +50,19 @@ paths:
                   type: integer
                   format: int64
                   description: The file size in bytes
+          multipart/form-data:
+            description: >
+              Legacy flow for backwards compatibility. Uploads the file directly
+              to the server in a single request.
+            schema:
+              type: object
+              required:
+                - file
+              properties:
+                file:
+                  type: string
+                  format: binary
+                  description: The file to upload
       responses:
         "200":
           description: File uploaded successfully
--- a/backend/internal/app/bootstrap/app.go
+++ b/backend/internal/app/bootstrap/app.go
@@ -292,9 +292,9 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 	mux.Handle("POST /api/v1/payments/webhook", http.HandlerFunc(paymentHandler.HandleWebhook))
 	mux.Handle("POST /api/v1/payments/validate", http.HandlerFunc(paymentHandler.HandleValidateReference))
 
-	// When using local storage, this endpoint serves the actual file bytes.
-	// It's made public since it's the equivalent of a presigned URL when using S3.
+	// When using local storage, these endpoints serve as mocks for S3.
 	if _, ok := storageDriver.(*drivers.LocalFSDriver); ok {
+		mux.HandleFunc("PUT /api/v1/uploads/local-put/{key}", uploadHandler.UploadContentLocal)
 		mux.HandleFunc("GET /api/v1/uploads/{key}/content", uploadHandler.DownloadContent)
 	}
 
--- a/backend/internal/app/bootstrap/app.go
+++ b/backend/internal/app/bootstrap/app.go
@@ -294,7 +294,7 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 
 	// When using local storage, these endpoints serve as mocks for S3.
 	if _, ok := storageDriver.(*drivers.LocalFSDriver); ok {
-		mux.HandleFunc("PUT /api/v1/uploads/local-put/{key}", uploadHandler.UploadContentLocal)
+		mux.HandleFunc("PUT /api/v1/uploads/{key}/content", uploadHandler.UploadContentLocal)
 		mux.HandleFunc("GET /api/v1/uploads/{key}/content", uploadHandler.DownloadContent)
 	}
 
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -62,6 +62,7 @@ type StorageConfig struct {
 	S3SecretKey    string
 	S3UseSSL       bool
 	S3PublicURL    string
+	LocalPutSecret string
 }
 
 type AuthConfig struct {
@@ -133,6 +134,7 @@ func Load() (*Config, error) {
 			S3SecretKey:    os.Getenv("STORAGE_S3_SECRET_KEY"),
 			S3UseSSL:       getBoolOrDefault("STORAGE_S3_USE_SSL", true),
 			S3PublicURL:    os.Getenv("STORAGE_S3_PUBLIC_URL"),
+			LocalPutSecret: getEnvOrDefault("STORAGE_LOCAL_PUT_SECRET", "local-dev-secret"),
 		},
 		Auth: AuthConfig{
 			JWKSURL:               getEnvOrDefault("AUTH_JWKS_URL", "https://localhost:8090/oauth2/jwks"),
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // Config holds all configuration for the application
@@ -63,6 +64,7 @@ type StorageConfig struct {
 	S3UseSSL       bool
 	S3PublicURL    string
 	LocalPutSecret string
+	PresignTTL     time.Duration
 }
 
 type AuthConfig struct {
@@ -135,6 +137,7 @@ func Load() (*Config, error) {
 			S3UseSSL:       getBoolOrDefault("STORAGE_S3_USE_SSL", true),
 			S3PublicURL:    os.Getenv("STORAGE_S3_PUBLIC_URL"),
 			LocalPutSecret: getEnvOrDefault("STORAGE_LOCAL_PUT_SECRET", "local-dev-secret"),
+			PresignTTL:     parseDurationOrDefault(getEnvOrDefault("STORAGE_PRESIGN_TTL", "15m"), 15*time.Minute),
 		},
 		Auth: AuthConfig{
 			JWKSURL:               getEnvOrDefault("AUTH_JWKS_URL", "https://localhost:8090/oauth2/jwks"),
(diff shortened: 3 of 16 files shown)


backend/internal/uploads/http_handler.go
```diff
@@ -55,47 +56,83 @@ func (h *HTTPHandler) Upload(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		writeJSONError(w, http.StatusBadRequest, "file is required")
+	// New Presigned URL generation flow (application/json)
+	var req struct {
+		Filename string `json:"filename"`
+		MimeType string `json:"mime_type"`
+		Size     int64  `json:"size"`
+	}
+
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
```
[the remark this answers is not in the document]
ginaxu1: added

api/openapi-uploads.yaml
```diff
@@ -29,19 +29,24 @@ paths:
       requestBody:
         required: true
         content:
-          multipart/form-data:
+          application/json:
```
Aravinda-HWK, line 32: In the `/upload` endpoint, you still support both `multipart/form-data` and `application/json`. Therefore, don't we want to support both content types in the `openapi-uploads.yaml` file.

backend/internal/config/config.go
```diff
@@ -123,6 +124,7 @@ func Load() (*Config, error) {
 			S3SecretKey:    os.Getenv("STORAGE_S3_SECRET_KEY"),
 			S3UseSSL:       getBoolOrDefault("STORAGE_S3_USE_SSL", true),
 			S3PublicURL:    os.Getenv("STORAGE_S3_PUBLIC_URL"),
+			LocalPutSecret: getEnvOrDefault("STORAGE_LOCAL_PUT_SECRET", "local-dev-secret"),
```
Aravinda-HWK, line 139: I think this variable should be mentioned in the `.env.example` file. Otherwise, we will miss setting this.

Aravinda-HWK (changes requested):

backend/internal/uploads/drivers/local_fs.go
Aravinda-HWK: `ctx` parameter is not used

backend/internal/uploads/drivers/local_fs.go
```diff
@@ -139,5 +143,49 @@ func (d *LocalFSDriver) GetDownloadURL(ctx context.Context, key string, ttl time
...
+
+	return fmt.Sprintf("%s/api/v1/uploads/%s/content?%s", d.PublicURL, key, v.Encode()), nil
+}
+
+// VerifyDownloadToken checks if a provided download token is valid and not expired.
+func (d *LocalFSDriver) VerifyDownloadToken(key, token string, expiresAt int64) bool {
+	return VerifyDownloadToken(key, token, d.secretKey, expiresAt)
+}
+
+func (d *LocalFSDriver) GetUploadURL(ctx context.Context, key string, ttl time.Duration, contentType string, maxSizeBytes int64) (string, error) {
```
Aravinda-HWK: `ctx` parameter is not used

backend/internal/uploads/drivers/local_fs.go
```diff
@@ -139,5 +143,49 @@ func (d *LocalFSDriver) GetDownloadURL(ctx context.Context, key string, ttl time
...
+	}
+	expiresAt := time.Now().Add(ttl).Unix()
+	token := GenerateDownloadToken(key, d.secretKey, expiresAt)
+
+	// Returns a URL with security token and expiration
+	v := url.Values{}
+	v.Set("token", token)
+	v.Set("expiresAt", strconv.FormatInt(expiresAt, 10))
+
+	return fmt.Sprintf("%s/api/v1/uploads/%s/content?%s", d.PublicURL, key, v.Encode()), nil
```
Aravinda-HWK, line 161: In the `GenerateDownloadToken` function, generate a token by encoding `secretKey` and `expiredAt` time. But after returning the token, in line 158, `v.Encode()` is done. I am not sure whether we need to encode the `expiresAt` field with the encoded token again.

backend/internal/uploads/drivers/token.go
Aravinda-HWK, line 1: Can we simplify this file using the functions below? Because I think that this file has duplication.

```
package drivers

import (
	"crypto/hmac"
	"crypto/sha256"
	"encoding/hex"
	"fmt"
	"strings"
)

// sign creates an HMAC-SHA256 signature from given parts
func sign(secret string, parts ...any) string {
	strParts := make([]string, len(parts))
	for i, p := range parts {
		strParts[i] = fmt.Sprintf("%v", p)
	}

	payload := strings.Join(strParts, "\x00")

	mac := hmac.New(sha256.New, []byte(secret))
	mac.Write([]byte(payload))

	return hex.EncodeToString(mac.Sum(nil))
}

// verify checks token validity using constant-time comparison
func verify(token, secret string, parts ...any) bool {
	if token == "" || secret == "" {
		return false
	}

	expected := sign(secret, parts...)
	return hmac.Equal([]byte(token), []byte(expected))
}
```

api/openapi-uploads.yaml
ginaxu1: updated

backend/internal/config/config.go
ginaxu1: added to `.env.example`

backend/internal/app/bootstrap/app.go
```diff
@@ -272,9 +272,9 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 	mux.Handle("POST /api/v1/payments/webhook", http.HandlerFunc(paymentHandler.HandleWebhook))
 	mux.Handle("POST /api/v1/payments/validate", http.HandlerFunc(paymentHandler.HandleValidateReference))
 
-	// When using local storage, this endpoint serves the actual file bytes.
-	// It's made public since it's the equivalent of a presigned URL when using S3.
+	// When using local storage, these endpoints serve as mocks for S3.
 	if _, ok := storageDriver.(*drivers.LocalFSDriver); ok {
+		mux.HandleFunc("PUT /api/v1/uploads/local-put/{key}", uploadHandler.UploadContentLocal)
```
mushrafmim: Shall we align with the same structure? with the Download URL below, so only the methods different.?

backend/internal/uploads/drivers/s3_driver.go
```diff
@@ -90,3 +90,21 @@ func (d *S3Driver) GetDownloadURL(ctx context.Context, key string, ttl time.Dura
 	}
 	return d.presignGet(ctx, key, ttl)
 }
+
+func (d *S3Driver) GetUploadURL(ctx context.Context, key string, ttl time.Duration, contentType string, maxSizeBytes int64) (string, error) {
+	if ttl == 0 {
+		ttl = 15 * time.Minute
```
mushrafmim: Should this be configurable?

backend/internal/uploads/drivers/local_fs.go
```diff
@@ -139,5 +143,49 @@ func (d *LocalFSDriver) GetDownloadURL(ctx context.Context, key string, ttl time
...
+	return VerifyDownloadToken(key, token, d.secretKey, expiresAt)
+}
+
+func (d *LocalFSDriver) GetUploadURL(ctx context.Context, key string, ttl time.Duration, contentType string, maxSizeBytes int64) (string, error) {
+	if d.PublicURL == "" {
+		return "", fmt.Errorf("public URL not configured for local storage")
+	}
+
+	if ttl == 0 {
+		ttl = 15 * time.Minute
```
mushrafmim: Should this be configurable?

mushrafmim (commented):
Only nits, otherwise the implementation looks perfect to me.

backend/internal/uploads/drivers/local_fs.go
ginaxu1: added `presignTTL` to the `LocalFSDriver` struct

backend/internal/uploads/drivers/s3_driver.go
ginaxu1: added `PresignTTL     time.Duration` to `StorageConfig` struct

backend/internal/app/bootstrap/app.go
ginaxu1: updated route definition to `PUT /api/v1/uploads/{key}/content`

mushrafmim: Nice!

backend/internal/uploads/drivers/s3_driver.go
mushrafmim: Nice!

backend/internal/uploads/drivers/local_fs.go
mushrafmim: Do we need ttl as the argument for each call, since we are configuring the instance level?

backend/internal/uploads/drivers/local_fs.go
```diff
@@ -139,5 +148,49 @@ func (d *LocalFSDriver) GetDownloadURL(ctx context.Context, key string, ttl time
...
+
+	return fmt.Sprintf("%s/api/v1/uploads/%s/content?%s", d.PublicURL, key, v.Encode()), nil
+}
+
+// VerifyDownloadToken checks if a provided download token is valid and not expired.
+func (d *LocalFSDriver) VerifyDownloadToken(key, token string, expiresAt int64) bool {
+	return VerifyDownloadToken(key, token, d.secretKey, expiresAt)
+}
+
+func (d *LocalFSDriver) GetUploadURL(ctx context.Context, key string, ttl time.Duration, contentType string, maxSizeBytes int64) (string, error) {
```
mushrafmim: Do we need ttl as the argument for each call, since we are configuring the instance level?

backend/internal/uploads/http_handler.go
```diff
@@ -55,47 +57,84 @@ func (h *HTTPHandler) Upload(w http.ResponseWriter, r *http.Request) {
...
-	// We avoid trusting client-supplied multipart headers for security and accuracy.
-	mimeType := header.Header.Get("Content-Type")
-	if seeker, ok := file.(io.ReadSeeker); ok {
-		// Sniff content type from the first 512 bytes.
-		sniffBuf := make([]byte, 512)
-		n, _ := seeker.Read(sniffBuf)
-		if n > 0 {
-			mimeType = http.DetectContentType(sniffBuf[:n])
-		}
+	if req.Filename == "" || req.MimeType == "" || req.Size <= 0 {
```
mushrafmim: Isn't this check too generic, I set the size to 0 and couldn't figure out what is the problem until I look at the code.

Isn't it better to have separate messages for each issue?

mushrafmim (changes requested):
I tested locally, It's working fine.

backend/internal/uploads/drivers/local_fs.go
sthanikan2000: +1

backend/internal/uploads/drivers/s3_driver.go
```diff
@@ -17,14 +17,19 @@ type S3Driver struct {
...
 	Bucket        string
 	PublicURL     string // Optional: Base URL if files are public
+	presignTTL    time.Duration
 }
 
-func NewS3Driver(client *s3.Client, bucket string, publicURL string) *S3Driver {
+func NewS3Driver(client *s3.Client, bucket string, publicURL string, presignTTL time.Duration) *S3Driver {
+	if presignTTL == 0 {
+		presignTTL = 15 * time.Minute
+	}
```
sthanikan2000, line 26: if we're defining default values, don't it need to come from default constants?

backend/internal/uploads/drivers/s3_driver.go
```diff
@@ -86,7 +91,25 @@ func (d *S3Driver) presignGet(ctx context.Context, key string, ttl time.Duration
 
 func (d *S3Driver) GetDownloadURL(ctx context.Context, key string, ttl time.Duration) (string, error) {
 	if ttl == 0 {
-		ttl = 15 * time.Minute
+		ttl = d.presignTTL
 	}
 	return d.presignGet(ctx, key, ttl)
 }
```
sthanikan2000: For the Upload there is a single function defining the GetUploadURL method. But for the GetDownloadURL you break down the function in to pieces. Shall we keep a consistency among the definition?

backend/internal/uploads/service.go
```diff
@@ -72,17 +66,45 @@ func (s *UploadService) Upload(ctx context.Context, filename string, reader io.R
...
+	if mime == "" {
+		mime = "application/octet-stream"
+	}
+	id := uuid.NewString()
+	ext := filepath.Ext(filename)
+	key := fmt.Sprintf("%s%s", id, ext)
+
+	// Generate a presigned URL for the upload.
+	// We use a 15-minute TTL by default.
+	uploadURL, err := s.Driver.GetUploadURL(ctx, key, 15*time.Minute, mime, size)
```
sthanikan2000: Don't hard code the TTL. Define a constant

backend/internal/uploads/service.go
```diff
@@ -72,17 +66,45 @@ func (s *UploadService) Upload(ctx context.Context, filename string, reader io.R
...
+
+// Upload handles the preparation of a file upload by generating a unique key
+// and a presigned/upload URL via the storage driver.
+func (s *UploadService) Upload(ctx context.Context, filename string, size int64, mime string) (*FileMetadata, error) {
+	if mime == "" {
+		mime = "application/octet-stream"
+	}
+	id := uuid.NewString()
+	ext := filepath.Ext(filename)
+	key := fmt.Sprintf("%s%s", id, ext)
```
sthanikan2000, line 89: According to this we're generating the key(filename), and call `s.Driver.GetUploadURL` to get the `d.PresignClient.PresignPutObject`. 
We can guarantee that s3 buckets will handle edge cases properly as,
1. First allocate the resource for the file
2. then return the put object URL

Is this handled properly for the `local_fs`?

mushrafmim (dismissed):

backend/internal/uploads/drivers/s3_driver.go
ginaxu1: added `const DefaultPresignTTL = 15 * time.Minute` to `backend/internal/uploads/drivers/constants.go`

backend/internal/uploads/service.go
ginaxu1: yeah current logic for LocalFSDriver correctly mimics S3 behavior
- check `LocalFSDriver.GetUploadURL` - it generates a signed token
- actual file created on disk only when the local-put handler calls `driver.Save`

backend/internal/uploads/drivers/s3_driver.go
ginaxu1: both methods follow the same pattern (call private helper functions `presignGet` and `presignPut` for the presigning logic)

backend/internal/uploads/service.go
```diff
@@ -72,17 +65,44 @@ func (s *UploadService) Upload(ctx context.Context, filename string, reader io.R
...
-	slog.InfoContext(ctx, "File uploaded successfully", "id", id, "key", key)
+	slog.InfoContext(ctx, "File uploaded successfully (legacy)", "id", id, "key", key)
+	return metadata, nil
+}
+
+// Upload handles the preparation of a file upload by generating a unique key
+// and a presigned/upload URL via the storage driver.
+func (s *UploadService) Upload(ctx context.Context, filename string, size int64, mime string) (*FileMetadata, error) {
+	if mime == "" {
+		mime = "application/octet-stream"
```
sthanikan2000: shall we add this as a constant

backend/internal/uploads/drivers/token.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"time"
+)
+
+// DefaultPresignTTL is the default time-to-live for presigned upload and download URLs.
+const DefaultPresignTTL = 15 * time.Minute
```
sthanikan2000:
```suggestion
// DefaultPresignTTL is the default time-to-live for presigned upload and download URLs.
const DefaultPresignTTL = 15 * time.Minute

// DefaultMinme
const DefaultMime = "application/octet-stream"
```

Shall we have it in the `backend/internal/uploads/models.go`? I think it is common to full upload service.

backend/internal/uploads/service.go
ginaxu1: added `const DefaultMimeType = "application/octet-stream"` to `backend/internal/uploads/drivers/constants.go`

sthanikan2000:
@ginaxu1 can you address the nits I commented? Then we don't need to touch this until we hit with new requirements

backend/internal/uploads/drivers/token.go
ginaxu1: better to have separate `backend/internal/uploads/drivers/constants.go` . moved the 2 constants there

sthanikan2000: yeah make sense

ginaxu1: `models.go` is in `package uploads`, if these 2 constants moved there, then the drivers package would need to import uploads to access those constants for its constructors (NewS3Driver, NewLocalFSDriver). this creates circular dependency uploads -> drivers -> uploads

sthanikan2000 (dismissed):
LGTM!

mushrafmim (dismissed):

Aravinda-HWK (dismissed):
LGTM

backend/internal/uploads/drivers/token.go
sthanikan2000: @ginaxu1 this comment make sense to me as well. Can you address this?

backend/internal/uploads/drivers/local_fs.go
ginaxu1: added _ prefix to ctx parameter. need to maintain the method signature to satisfy the StorageDriver interface

backend/internal/uploads/drivers/local_fs.go
ginaxu1: added _ prefix to ctx parameter. need to maintain the method signature to satisfy the StorageDriver interface

backend/internal/uploads/drivers/local_fs.go
ginaxu1: no need to encode again; just need `expiresAt` in query params for 2 server checks
1. expiration check: if link is past its TTL 
2. to verify the token: recalculate the signature and compare it against the provided token

backend/internal/uploads/drivers/token.go
ginaxu1: simplified

sthanikan2000 (approved):
LGTM!

Aravinda-HWK (approved):

PR #378, opened 2026-04-20
Phase 2: Integrate pre-signed URLs to OGA backend

ginaxu1:
## Summary
Building on the backend updates in #361, this PR integrates the OGA backend to fully proxy and fetch pre-signed upload operations. OGA backend can now correctly consume the secure upload URLs with the central NSW platform

## Changes
- Service integration: add `CreateUploadURL` into `ogaService` to securely proxy file metadata payloads and retrieve valid tokens natively.
- Add `HandleCreateUpload` HTTP endpoint in the OGA router mapped to `POST /api/oga/uploads` to handle direct client integrations for OGA.
- Unit Tests: add robust handler (`handler_test.go`) and service (`service_test.go`) test suites enforcing response integrity and edge case handling around the new proxies.

## Testing
- Pre-signed interaction contracts verified against the core NSW upload mechanisms


oga/internal/handler.go
```diff
@@ -247,3 +247,26 @@ func (h *OGAHandler) HandleGetUploadURL(w http.ResponseWriter, r *http.Request)
 		"expires_at":   time.Now().Add(15 * time.Minute).Unix(),
 	})
 }
+
+// HandleCreateUpload prepares an upload by requesting an upload URL from the main backend
+func (h *OGAHandler) HandleCreateUpload(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
ginaxu1: added TODO to add auth middleware; better to do later after #361 merge

oga/internal/constants.go
```diff
@@ -0,0 +1,14 @@
+package internal
+
+import "time"
+
+const (
+	// DefaultPresignTTL is the default time-to-live for presigned upload and download URLs
+	DefaultPresignTTL = 15 * time.Minute
```
sthanikan2000: we'll not specify this in the OGA side. This is up to the upload service to define the TTL for it self.

oga/internal/constants.go
```diff
@@ -0,0 +1,14 @@
...
+
+	// Application Statuses
+	StatusPending           = "PENDING"
+	StatusApproved          = "APPROVED"
+	StatusRejected          = "REJECTED"
+	StatusFeedbackRequested = "FEEDBACK_REQUESTED"
```
sthanikan2000: Changes related to refactor hard coded values to constants should be done as separate PR. Because this is not related to problem you're solving here

oga/internal/handler.go
```diff
@@ -244,6 +244,34 @@ func (h *OGAHandler) HandleGetUploadURL(w http.ResponseWriter, r *http.Request)
 
 	WriteJSONResponse(w, http.StatusOK, map[string]any{
 		"download_url": downloadURL,
-		"expires_at":   time.Now().Add(15 * time.Minute).Unix(),
+		"expires_at":   time.Now().Add(DefaultPresignTTL).Unix(),
```
sthanikan2000: this should come from `h.service.GetDownloadURL(r.Context(), key)`. If you correctly define the `struct` of the response form the upload service, you can get the required details properly.

oga/internal/handler.go
sthanikan2000: we can't enforce a pattern for the `key` in the OGA side. It is up to the `upload` service to decide which pattern to use.

oga/internal/service.go
```diff
@@ -333,12 +340,6 @@ func (s *ogaService) ReviewApplication(ctx context.Context, taskID string, revie
 		return fmt.Errorf("failed to send response to service: %w", err)
 	}
 
-	if err := s.store.UpdateStatus(taskID, status, reviewerResponse); err != nil {
-		// TODO: If this fails, we have already sent the response to the service but failed to update our record of it. We should consider how to handle this edge case - for now we just log an error.
-		slog.ErrorContext(ctx, "failed to update application status in database after successful service call", "taskID", taskID, "status", status, "error", err)
-		return fmt.Errorf("failed to update application status in database after successful service call: %w", err)
-	}
-
```
sthanikan2000: you resolved the rebase conflict wrong. you're reverting the changes I did in the https://github.com/OpenNSW/nsw/pull/354/changes

oga/internal/service.go
```diff
@@ -384,12 +389,6 @@ func (s *ogaService) FeedbackApplication(ctx context.Context, taskID string, con
 		return fmt.Errorf("failed to send feedback to service: %w", err)
 	}
 
-	if err := s.store.AppendFeedback(taskID, entryMap); err != nil {
-		// TODO: If this fails, we have already sent the feedback to the service but failed to update our record of it. We should consider how to handle this edge case - for now we just log an error.
-		slog.ErrorContext(ctx, "failed to store feedback in database after successful service call", "taskID", taskID, "round", entry.Round, "error", err)
-		return fmt.Errorf("failed to store feedback in database after successful service call: %w", err)
-	}
-
```
sthanikan2000: you resolved the rebase conflict wrong. you're reverting the changes I did in the https://github.com/OpenNSW/nsw/pull/354/changes

sthanikan2000 (changes requested):

oga/internal/constants.go
ginaxu1: removed the constants file

oga/internal/service.go
ginaxu1: Reverted this file to `origin/main` to fix rebase errors and re-added only the necessary parts `CreateUploadURL` and update to `GetDownloadURL` (change return type to return both the URL and the expiration timestamp)

oga/internal/handler.go
ginaxu1: removed `storageKeyRx` and the regex validation here in `HandleGetUploadURL`

oga/internal/handler.go
ginaxu1: updated `HandleGetUploadURL` to use the expires_at returned by the service instead of calculating it locally

oga/internal/constants.go
ginaxu1: removed all references to `DefaultPresignTTL` in OGA side

Aravinda-HWK (dismissed):

oga/internal/service.go
```diff
@@ -396,33 +399,58 @@ func (s *ogaService) FeedbackApplication(ctx context.Context, taskID string, con
...
 
 	if metadata.DownloadURL == "" {
-		return "", fmt.Errorf("metadata response missing download_url")
+		return "", 0, fmt.Errorf("metadata response missing download_url")
 	}
 
 	slog.InfoContext(ctx, "resolved download URL from metadata", "key", key, "downloadURL", metadata.DownloadURL)
-	return metadata.DownloadURL, nil
+	return metadata.DownloadURL, metadata.ExpiresAt, nil
+}
```
mushrafmim: It is a go best practice to return (object, err) only two values. Instead of returning  3, better to directly return the `metadata`.

oga/internal/handler.go
```diff
@@ -244,6 +235,34 @@ func (h *OGAHandler) HandleGetUploadURL(w http.ResponseWriter, r *http.Request)
 
 	WriteJSONResponse(w, http.StatusOK, map[string]any{
 		"download_url": downloadURL,
-		"expires_at":   time.Now().Add(15 * time.Minute).Unix(),
+		"expires_at":   expiresAt,
 	})
```
mushrafmim: If you modify the function according to this [comment](https://github.com/OpenNSW/nsw/pull/378/changes#r3117457030) can easily send the same response via this endpoint, instead of rebuilding it.

oga/internal/service.go
```diff
@@ -396,33 +399,58 @@ func (s *ogaService) FeedbackApplication(ctx context.Context, taskID string, con
...
+
+// CreateUploadURL proxies an upload initialization request to the main backend.
+func (s *ogaService) CreateUploadURL(ctx context.Context, payload []byte) (map[string]any, error) {
```
mushrafmim, line 430: The CreateUploadURL shouldn't blindly accept any `payload []byte` and forward it to the `nsw`, instead it should have the same type checking and then forwarding.

mushrafmim (changes requested):

oga/internal/service.go
ginaxu1: thanks, updated `GetDownloadURL` to `(map[string]any, error)`

oga/internal/service.go
ginaxu1: updated CreateUploadURL to parse the []byte payload into a struct, checking that filename, mime_type, and size are valid

oga/internal/handler.go
ginaxu1: updated to `WriteJSONResponse(w, http.StatusOK, metadata)`

oga/internal/handler.go
```diff
@@ -230,20 +226,40 @@ func (h *OGAHandler) HandleGetUploadURL(w http.ResponseWriter, r *http.Request)
...
+	// TODO: Add Authentication & Authorization middleware
+	// Access must be restricted to authorized OGA officers to prevent unauthorized users
+	// from generating proxy pre-signed upload URLs. Introduce a configuration flag (e.g. OGA_DISABLE_AUTH)
+	// to make bypassing explicit for specific environments
+	if r.Method != http.MethodPost {
+		WriteJSONError(w, http.StatusMethodNotAllowed, "Method not allowed")
+		return
+	}
+
+	r.Body = http.MaxBytesReader(w, r.Body, 1<<20) // Limit to 1MB
```
sthanikan2000: Can't we make this configurable? And have a default value 32MB

sthanikan2000 (dismissed):

oga/internal/handler.go
ginaxu1: added `MaxRequestBytes int64` to HTTPHandler

oga/cmd/server/main.go
```diff
@@ -66,6 +66,7 @@ func main() {
 
 	// Initialize handlers
 	handler := internal.NewOGAHandler(service)
+	handler.MaxRequestBytes = cfg.MaxRequestBytes
```
mushrafmim: Shouldn't this be passed via the NewOGAHandler() call? instead of directly setting that value. What if they set an invalid value? like `-1`

oga/internal/config.go
```diff
@@ -76,6 +77,12 @@ func LoadConfig() (Config, error) {
 		},
 	}
 
+	maxRequestBytes, err := parseInt64Env("OGA_MAX_REQUEST_BYTES", 32<<20)
+	if err != nil {
+		return Config{}, err
+	}
+	cfg.MaxRequestBytes = maxRequestBytes
+
 	tokenInsecureSkipVerify, err := parseBoolEnv("OGA_NSW_TOKEN_INSECURE_SKIP_VERIFY", false)
```
mushrafmim, line 86: This validation should be done inside the NewHandler() call, related to the comment on `main.go`

mushrafmim (changes requested):

oga/internal/config.go
ginaxu1: moved check to inside `NewOGAHandler`

oga/cmd/server/main.go
ginaxu1: updated `func NewOGAHandler(service OGAService, maxRequestBytes int64) (*OGAHandler, error) {`

mushrafmim (approved):

Aravinda-HWK (approved):

PR #390, opened 2026-04-22
feat: refactor notifications package with providers

### LINKED ISSUE #367: [Backend][Notifications] Refactor notification package.
### Current Limitation

Address: 
- [Graceful shutdown Handling](https://github.com/OpenNSW/nsw/pull/360#pullrequestreview-4126562376)
- [Better Config Handling](https://github.com/OpenNSW/nsw/pull/360#discussion_r3098056519)
- Refactoring to make notification more easier to use.

### Suggested Improvement

Address these issues.

### Version

_No response_

### Additional Context

_No response_

maneeshaxyz:
## Summary

This PR introduces structured request types, provider-based delivery routing, and template rendering for both channels.

These changes are needed to replace the current notification flows with a single, extensible package that standardizes message delivery and simplifies future maintenance.

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [X] Refactoring (no functional changes)
- [ ] Performance improvement

## Changes Made

- Implemented pkg/notifications to handle routing and sending notifications via email and SMS.
- Added core types: Request, EmailRequest, and SMSRequest for structured notification data.
- Created Provider interface for different notification channels.
- Extended Manager to manage providers and template rendering.
- Introduced email and SMS providers with external service integration.
- Fixed template rendering for email and SMS notifications.

## Testing

- [x] All existing tests pass

## Checklist

## Related Issues

Closes #367 

## Screenshots/Demo

N/A

## Additional Notes

Large amount of file changes due to notification -> notifications, most logic to be reviewed is in the Providers and the `notifications.go`

## Deployment Notes


ginaxu1:
some considerations
1. @mushrafmim feel free to add your thoughts: better split the `Request` struct into specific EmailRequest and SMSRequest types, otherwise seems antipattern to have SMS provider silently ignore fields it doesn't care about ( `Subject` and `HTMLBody` ) and likewise with Email
2. check if template.go uses an in-memory cache (sync.Map or pre-parsed template tree) rather than reading from disk per request
3. make sure no long-running network calls (SMTP/GovSMS) are inadvertently bound to cancelling HTTP contexts

maneeshaxyz:
@ginaxu1 @mushrafmim 
>     1. @mushrafmim feel free to add your thoughts: better split the `Request` struct into specific EmailRequest and SMSRequest types, otherwise seems antipattern to have SMS provider silently ignore fields it doesn't care about ( `Subject` and `HTMLBody` ) and likewise with Email

The internal Request type is a transport envelope for the Provider interface. Public callers use EmailRequest/SMSRequest. The Manager maps those to Request before dispatch, ensuring each provider only receives fields relevant to its channel.  

>     2. check if template.go uses an in-memory cache (sync.Map or pre-parsed template tree) rather than reading from disk per request.

It does. templateCache uses sync.RWMutex + plain maps 

>     3. make sure no long-running network calls (SMTP/GovSMS) are inadvertently bound to cancelling HTTP contexts

Covered. Both SendEmail and SendSMS use context.WithoutCancel(ctx) inside the goroutine, so the parent HTTP    request context cancelling doesn't abort the in-flight network call

backend/internal/app/bootstrap/app.go
```diff
@@ -189,21 +190,26 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
...
+	})
+
+	notificationManager := notifications.New(
+		notifications.Config{
+			EmailTemplateRoot: cfg.Notification.EmailTemplateRoot,
+			SMSTemplateRoot:   cfg.Notification.SMSTemplateRoot,
+		},
+		emailProvider,
+		govsmsProvider,
+	)
```
sthanikan2000:
```suggestion
	emailProvider := email.NewService(email.ServiceConfig{
		BaseURL: cfg.Notification.EmailServiceURL,
		Token:   cfg.Notification.EmailServiceToken,
    })

	smsProvider := sms.NewGovSMS(sms.GovSMSConfig{
		BaseURL:  cfg.Notification.GovSMSBaseURL,
		UserName: cfg.Notification.GovSMSUsername,
		Password: cfg.Notification.GovSMSPassword,
		SIDCode:  cfg.Notification.GovSMSSIDCode,
	})

	notificationManager := notifications.NewManager()
	notificationManager.RegisterProvider(emailProvider,cfg.Notification.EmailTemplateRoot)
	notificationManager.RegisterProvider(smsProvider,cfg.Notification.SMSTemplateRoot)
```

Can we wire the notification channels like this?

sthanikan2000 (changes requested):

sthanikan2000 (commented):
The way you implemented the `providers` is fine. But need some improvement in the `NotificationManager` and the `app.go` wiring.
Will share a doc with you related to proper wiring

sthanikan2000 (changes requested):
The interfaces and implementations should similar to this:
- The core files of this package should be `notification.go`, `provider.go`, `providers/email/service.go`, `providers/sms/service.go`

- `notification.go`: manager orchestrate multiple notification providers registered with it.
```go
type ChannelType string

const (
	ChannelSMS   ChannelType = "sms"
	ChannelEmail ChannelType = "email"
)

type NotificationConfig struct {
	ConfigPath string
}

type NotificationManager struct {
	providers  map[ChannelType]Provider

	// userProfileService
}

func NewNotificationManager(config NotificationConfig, providers []NotificationProvider) *NotificationManger {
	// Load the `configs` from the file at `ConfigPath`, using a helper function that can parse `json` or `yaml` file
	
	// for provider in provider:
	// 		provider.SetAndValidateConfig(configs[provider.Type()])
	// 		provider[provider.Type()] = provider
}

func (nm *NotificationManager) SendNotification(context.Context, payload any) error {
	// Check for the notification preference from the `userProfileService`

	// Call the `provider.Send` accordingly`
}
```

- `provider. go`: interface definition for any provider. To be a part of the Notification channel, a provider should implement the interface defined there.
```go
// Provider is the contract every notification channel must satisfy.
// Adding a new channel = implement this interface, then register it.
type NotificationProvider interface {
	Type() ChannelType
	SetAndValidateConfig(config any) error // Validation will be happen here.
	Send(ctx context.Context, req any) error
}
```

- Each providers should implement the `Provider` interface

- finally `app.go` wiring will be as follows:
```go
smsProvider := sms.NewSMSProvider()
emailProvider := email.NewEmailProvider()

providers := { smsProvider, emailProvider }

nm := notification.NewNotificationManager(config, providers,httpClient)
```

> [!NOTE]
> I may missed best practices do a research before implementation. Overall Design should be this.
>
> Problem you're solving here have similarity with PR #414. Refer this review comment as well: https://github.com/OpenNSW/nsw/pull/414#pullrequestreview-4202282027

**CC: @mushrafmim check for the correctness**

sthanikan2000:
@maneeshaxyz  updated my previous comment. Take a look.

backend/internal/app/bootstrap/app.go
```diff
@@ -133,21 +133,13 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
...
-	notificationManager.RegisterEmailChannel(emailChannel)
-
-	// TODO: Add SMS channel if needed
-	// smsChannel := channels.NewSMSChannel(...)
-	// notificationManager.RegisterSMSChannel(smsChannel)
+	var notificationManager *notification.Manager
+	if cfg.Notification.ConfigPath != "" {
+		notificationManager, err = notification.NewManager(cfg.Notification.ConfigPath)
+		if err != nil {
+			log.Printf("notifications init failed, continuing without notifications: %v", err)
```
sthanikan2000: use `slog.Error` or `slog.WARN`

backend/internal/config/config.go
```diff
@@ -45,12 +43,7 @@ type CORSConfig struct {
...
 
 type NotificationConfig struct {
-	SMTPHost     string
-	SMTPPort     int
-	SMTPUsername string
-	SMTPPassword string
-	SMTPSender   string
-	TemplateRoot string
+	ConfigPath string
 }
```
sthanikan2000: let's define this in the `pkg/notification/config.go` as `Config`
Check how we're importing for other services and packages

backend/internal/config/config.go
```diff
@@ -113,17 +106,7 @@ func Load() (*Config, error) {
 			InsecureSkipTLSVerify: getBoolOrDefault("AUTH_JWKS_INSECURE_SKIP_VERIFY", false),
 		},
 		Notification: NotificationConfig{
```
sthanikan2000:
```go
Notification: notification.Config{
     ConfigPath: getEnvOrDefault("NOTIFICATIONS_CONFIG_PATH", "configs/notification.json")
}
```

backend/pkg/notification/testdata/email/otp.tmpl
sthanikan2000, line 1: What is the reason for the changes in this file?

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,111 @@ package notification
...
 
-// EmailChannel defines the interface for an email notification provider.
-type EmailChannel interface {
-	Send(ctx context.Context, payload EmailPayload) map[string]error
+func defaultProviders(client *http.Client) []core.Provider {
+	return []core.Provider{
+		email.NewProvider(client),
+		sms.NewProvider(client),
+	}
 }
```
sthanikan2000: I don't think we need default providers. Providers without the correct config is no use right?

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,111 @@ package notification
...
 	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"regexp"
+
+	"github.com/OpenNSW/nsw/pkg/notification/internal/core"
+	"github.com/OpenNSW/nsw/pkg/notification/providers/email"
+	"github.com/OpenNSW/nsw/pkg/notification/providers/sms"
```
sthanikan2000: `NotificationManger` should not know what are the provider available. It should be dynamic.

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,111 @@ package notification
...
+func (m *Manager) Send(ctx context.Context, req Request) error {
+	if err := req.Validate(); err != nil {
+		return err
+	}
+	p, ok := m.providers[req.Channel]
+	if !ok {
+		return fmt.Errorf("unsupported channel: %s", req.Channel)
+	}
+	return p.Send(ctx, req)
 }
```
sthanikan2000, line 50: Shouldn't this logic will be a loop?
Are you assuming we need to send notification to one channel?

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,111 @@ package notification
...
 
-// EmailPayload contains email-specific notification data.
-type EmailPayload struct {
-	BasePayload
-	Recipients []string
-	Subject    string
-	Body       string // Used if TemplateID is empty
+func NewManager(configPath string) (*Manager, error) {
+	return newManager(configPath, nil, nil)
 }
```
sthanikan2000:
```suggestion
func NewManager(configPath string, providers []core.Provider) (*Manager, error) {
	return newManager(configPath, providers)
}
```

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,111 @@ package notification
...
+	expanded, err := expandEnv(raw)
+	if err != nil {
+		return nil, err
+	}
+	var cfgMap map[string]json.RawMessage
+	if err := json.Unmarshal(expanded, &cfgMap); err != nil {
+		return nil, fmt.Errorf("parse notification config: %w", err)
+	}
+	return cfgMap, nil
 }
```
sthanikan2000: Move this to a separate file.

backend/internal/app/bootstrap/app.go
```diff
@@ -133,21 +133,13 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
...
-		TemplateRoot: cfg.Notification.TemplateRoot,
-	})
-	notificationManager.RegisterEmailChannel(emailChannel)
-
-	// TODO: Add SMS channel if needed
-	// smsChannel := channels.NewSMSChannel(...)
-	// notificationManager.RegisterSMSChannel(smsChannel)
+	var notificationManager *notification.Manager
+	if cfg.Notification.ConfigPath != "" {
+		notificationManager, err = notification.NewManager(cfg.Notification.ConfigPath)
```
sthanikan2000:
```suggestion
        smsProvider := sms.NewProvider(httpClient)
        emailProvider := email.NewProvider(httpClient)
		notificationManager, err = notification.NewManager(
		    cfg.Notification.ConfigPath,
		    &{ smsProvider, emailProvider) // Check the correct way to define this Array
		}
```

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,111 @@ package notification
...
+func NewManager(configPath string) (*Manager, error) {
+	return newManager(configPath, nil, nil)
 }
 
-// SMSPayload contains phone-based notification data (SMS, WhatsApp, etc.).
-type SMSPayload struct {
-	BasePayload
-	Recipients []string
-	Body       string // Used if TemplateID is empty
+func newManager(configPath string, providers []core.Provider, httpClient *http.Client) (*Manager, error) {
```
sthanikan2000:
```suggestion
func newManager(configPath string, providers []core.Provider) (*Manager, error) {
```

sthanikan2000 (changes requested):
There are some issue in the wiring part. Let's do it offline together. Implementation is fine. There should be small refactor needs to be done.

backend/pkg/notification/notification.go
sthanikan2000: Added TODO for what need to be done here

backend/pkg/notification/testdata/email/otp.tmpl
sthanikan2000: @maneeshaxyz I need you help on this

backend/pkg/notification/loader.go
```diff
@@ -0,0 +1,81 @@
...
+// If any environment variables referenced in the data are not set, expandEnv returns
+// an error listing all missing variables and leaves those placeholders unchanged.
+//
+// Parameters:
+//   - data: The byte slice containing potential environment variable placeholders
+//
+// Returns:
+//   - A byte slice with all environment variables expanded
+//   - An error if any referenced environment variables are not set
+func expandEnv(data []byte) ([]byte, error) {
```
ginaxu1, line 12: rewrite this function. 

Go best practice is to load the JSON into a struct first, and use a Go library to override struct fields with environment variables, rather than regex-replacing raw JSON bytes. 

If you absolutely have to keep regex, at least handle the error and avoid blind slicing
```
encoded, err := json.Marshal(val)
if err != nil {
    // Log any errors, don't ignore 
    return match
}
// use trim, safer 
str := string(encoded)
str = strings.TrimPrefix(str, `"`)
str = strings.TrimSuffix(str, `"`)
return []byte(str)
```

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,49 @@ package notification
...
 
-// EmailConfig holds configuration for email channel.
-type EmailConfig struct {
-	SMTPHost     string
-	SMTPPort     int
-	SMTPUsername string
-	SMTPPassword string
-	SMTPSender   string
-	TemplateRoot string // Directory containing email templates
+	// TODO: req can contain multiple channels, so we should send to all of them
```
ginaxu1, line 42: why not update `Request` struct in this PR to have `Channels []ChannelType` then??

ginaxu1 (changes requested):

backend/pkg/notification/notification.go
```diff
@@ -2,46 +2,49 @@ package notification
...
 }
 
-// EmailPayload contains email-specific notification data.
-type EmailPayload struct {
-	BasePayload
-	Recipients []string
-	Subject    string
-	Body       string // Used if TemplateID is empty
+func NewManager(configPath string, providers []Provider) (*Manager, error) {
+	return newManager(configPath, providers)
```
ginaxu1, line 13: what's the point of this? why not remove `newManager` entirely and put the logic directly inside NewManager

ginaxu1 (changes requested):

backend/pkg/notification/config.go
```diff
@@ -0,0 +1,14 @@
+package notification
+
+import "fmt"
+
+type Config struct {
+	Path string
+}
+
+func (c *Config) Validate() error {
```
ginaxu1: why need the pointer receiver here? risks nil-pointer panics if called on an uninitialized pointer

Better use a value receiver for small structs.

```
func (c Config) Validate() error { ... 
```

backend/pkg/notification/providers/email/service.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type emailPayload struct {
+	To       string `json:"to"`
+	Subject  string `json:"subject"`
+	TextBody string `json:"text_body"`
+	HTMLBody string `json:"html_body,omitempty"`
+}
+
+func (p *emailProvider) Send(ctx context.Context, req notification.Request) error {
```
ginaxu1, line 55: nit: duplicate HTTP logic in `providers/email/service.go` + `providers/sms/service.go` 

consider writing shared HTTP helper

backend/pkg/notification/providers/email/service.go
```diff
@@ -0,0 +1,98 @@
...
+		Subject:  req.Subject,
+		TextBody: req.Body,
+		HTMLBody: req.HTMLBody,
+	}
+	body, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("marshal email payload: %w", err)
+	}
+
+	endpoint, err := url.JoinPath(p.cfg.BaseURL, "emails")
```
ginaxu1, line 67: noticed you append "emails" but for SMS you don't append (p.cfg.BaseURL is used as-is) double check this is correct

ginaxu1 (changes requested):

backend/pkg/notification/notification.go
sthanikan2000: That's going to be done in a separate PR. Lot of thinks needs to be addressed

backend/pkg/notification/config.go
sthanikan2000: Good catch Regina,
I gone through other configs as well and figured out the same issue is there in the other config files as well. I created the issue #453 for it. Also let's fix the notification config in this PR.

backend/pkg/notification/loader.go
sthanikan2000: The suggestion to load JSON into a struct first and use a library for env overrides isn't applicable here, the config is a map[string]json.RawMessage with unknown provider shapes at this layer, so there's no struct to map to. Template substitution before parsing is the right approach.

That said, the concerns about blind slicing and the unhandled error were valid. Replaced the custom regex with stdlib os.Expand, which natively handles ${VAR} syntax and eliminates the match-byte slicing entirely. The json.Marshal error is still discarded, but with an explicit comment explaining why it's safe, marshaling a string in Go never errors.

backend/pkg/notification/notification.go
sthanikan2000: newManager is kept separate because tests in the same package call it directly - this gives them a stable handle to the raw constructor independent of any future logic added to NewManager. Collapsing them would either break the tests or force them to go through production-only setup they shouldn't need.

backend/pkg/notification/providers/email/service.go
sthanikan2000: Noted.
@maneeshaxyz Let's address this when we implement the exact logic for the provider services in the follow up PR.

backend/pkg/notification/providers/email/service.go
sthanikan2000: @maneeshaxyz can you check with this

PR #394, opened 2026-04-23
feat(auth,user-profile): establish user record foundation with phone number support

sthanikan2000:
## Summary

This PR establishes the backend foundation for UserProfile management and notification-ready user data. It separates persisted user data from runtime auth context, adds phone number support through the auth flow, and updates the IdP bootstrap so new users can carry a phone number from the start.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Other (please describe):

## Changes Made

- Introduced `UserRecord` as the persisted auth/user model and kept `UserContext` for runtime request context.
- Updated auth service flows to read and write persisted user data through `UserRecord`.
- Added `phone_number` support to:
  - token claims
  - principal mapping
  - middleware context creation
  - user upsert payloads
- Updated auth context handling to carry roles alongside user identity fields.
- Renamed the backing database table to `user_records` and updated the initial schema migration accordingly.
- Updated IdP sample bootstrap resources to support `phone_number` for seeded users and application attributes.
- Updated auth tests to match the simplified auth model and to cover the key parser, service, and middleware paths.

## Testing

- [x] I have tested this change locally
- [x] I have added unit tests for new functionality
- [x] I have tested edge cases
- [x] All existing tests pass

Validation:
- `go test ./internal/auth -cover`
- Coverage for auth is 80.2%

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have checked that there are no merge conflicts

## Related Issues

No specific issues created

## Screenshots/Demo

N/A

## Additional Notes

This PR is the foundation for upcoming UserProfile management work and enables phone number data to flow through the auth and user context layers for later notification use cases.

## Deployment Notes

Apply the backend schema migration for `user_records` and ensure IdP bootstrap resources are updated before seeding users with phone numbers.

If you want, I can also turn this into a shorter GitHub-ready version or make the tone more technical/reviewer-focused.

PR #435, opened 2026-05-03
feat(auth,user-profile): separate persisted user profiles from auth and add phone number support

### LINKED ISSUE #434: [NSW Backend][Auth, Profiles] User Profile Service
## Problem

The current auth flow is doing more than authentication. It is also carrying persisted user data inside request-level auth context, which creates unnecessary preload overhead on every request and makes the auth package harder to reuse in other systems.

At the same time, several services need centralized user data on demand rather than on every request. For example, a notification service may need a user's email address and phone number only when it is actually sending a message. Keeping that data tied to auth context makes the design less flexible and harder to scale across multiple services.

We need a generalized auth layer and a separate user profile service so that centralized user data can be fetched only when a service needs it.

## Proposed Solution

Introduce a dedicated User Profile Service that owns persisted user data and exposes it to other services on demand.

The auth package should remain focused on token parsing, principal mapping, and request authentication only. It should not preload stored user data into every request. Instead, services that need user-centric data should call the user profile service directly when needed.

Expected responsibilities:

- Auth should validate tokens and build runtime request context.
- User Profile Service should store and manage centralized user data such as:
  - user ID
  - email
  - phone number
  - organization or tenant identity
  - notification preferences
- Other services should fetch user data on demand from the user profile service.

This will make the architecture easier to reuse across services such as notification, workflow, and future integrations that need user-centric metadata.

## Additional Context

This feature is intended to support a more generalized architecture for NSW:

- Auth should be reusable without depending on user persistence.
- User profile data should be centralized for services that need it.
- Request handling should avoid unnecessary data loading when a service does not need user details.
- Notification and similar services should be able to request user data only when required.

sthanikan2000:
## Summary

This PR combines the user-profile foundation from PR #394 with the auth refactor that splits persisted user data out of the auth package. The result is a reusable auth layer that handles token parsing and request context only, while user record persistence moves into `internal/profile/user`.

It also preserves the phone number flow end-to-end so new users can carry notification-ready data from the IdP into persisted user records.

## Type of Change

## Changes Made

- Split persisted user data into `internal/profile/user` and kept `internal/auth` focused on authentication and request context.
- Introduced `UserProfileService` as an optional dependency for auth, so auth can run without user persistence.
- Kept `tokenExtractor` required in `auth.Manager` and made the constructor enforce that contract.
- Added `phone_number` support through token claims, principal mapping, auth context creation, and user record persistence.
- Renamed the persisted table to `user_records` and updated the initial schema migration.
- Updated middleware and workflow routers to stop relying on preloaded user context and instead pass user metadata on demand.
- Updated IdP bootstrap resources so seeded users and app attributes can carry phone numbers.
- Refreshed auth tests and added coverage for the new manager contract and generic auth behavior.

## Testing

- [x] All existing tests pass

Validation:
- `cd backend && go test ./...`

## Checklist

## Related Issues

Closes #434 

## Additional Notes

This PR intentionally merges the auth decoupling work with the user-profile foundation so the final result is a single coherent backend change instead of two partially overlapping PRs.

## Deployment Notes

Apply the backend schema migration for `user_records` and ensure the updated IdP bootstrap resources are used before seeding users.


backend/internal/auth/token_parser.go
```diff
@@ -219,17 +221,18 @@ func (te *TokenExtractor) userPrincipalFromClaims(claims *tokenClaims) (*UserPri
...
 	}
-	if claims.OUHandle == nil {
-		return nil, fmt.Errorf("jwt missing ouHandle claim for user principal")
-	}
 	if claims.OUID == nil {
 		return nil, fmt.Errorf("jwt missing ouId claim for user principal")
 	}
+	if len(claims.Roles) == 0 {
+		return nil, fmt.Errorf("jwt missing roles claim for user principal")
+	}
```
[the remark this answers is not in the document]
sthanikan2000: Good catch. An Authn middleware should not reject this if `Roles` not there. AuthZ middleware will do this check and returns `StatusForbidden` if there is no access.

backend/internal/auth/middleware.go
```diff
@@ -73,41 +69,39 @@ func Middleware(authService *AuthService, tokenExtractor *TokenExtractor) func(h
...
+						principal.UserPrincipal.UserID,
+						principal.UserPrincipal.Email,
+						func() string {
+							if principal.UserPrincipal.PhoneNumber != nil {
+								return *principal.UserPrincipal.PhoneNumber
+							}
+							return ""
+						}(),
+						principal.UserPrincipal.OUID,
+					)
```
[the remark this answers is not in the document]
sthanikan2000: Addressed by adding `UserExists` function to the `UserProfileService`

backend/internal/profile/user/service.go
```diff
@@ -0,0 +1,123 @@
...
+		Email:       email,
+		PhoneNumber: phone,
+		OUID:        ouID,
+		NSWData:     []byte(`{}`),
+	}
+
+	result := s.db.Create(record)
+	if result.Error != nil {
+		// Check for unique constraint violation (user already exists)
+		if result.Error.Error() == "ERROR: duplicate key value violates unique constraint \"user_records_pkey\"" {
```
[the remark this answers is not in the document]
sthanikan2000: Upstream service which using the `UserProfileService` should check the existence of record before making call to create it. If it making create request directly and there is an duplicate record, this method will nit return any special error.

backend/internal/profile/user/service.go
```diff
@@ -0,0 +1,144 @@
...
+	record := &Record{
+		UserID:      userID,
+		Email:       email,
+		PhoneNumber: phone,
+		OUID:        ouID,
+		NSWData:     []byte(`{}`),
+	}
+
+	result := s.db.Create(record)
+	if result.Error != nil {
```
ginaxu1, line 94: `CreateUser` not idempotent, check again. best to use GORM's `ON CONFLICT DO NOTHING`

ginaxu1 (changes requested):

backend/internal/auth/middleware.go
```diff
@@ -73,41 +69,69 @@ func Middleware(authService *AuthService, tokenExtractor *TokenExtractor) func(h
...
-						OUHandle: principal.UserPrincipal.OUHandle,
-						OUID:     principal.UserPrincipal.OUID,
-						NSWData:  []byte(`{}`),
+						// user does not exist - attempt to create
+						create_err := userProfileService.CreateUser(
+							principal.UserPrincipal.UserID,
+							principal.UserPrincipal.Email,
+							func() string {
+								if principal.UserPrincipal.PhoneNumber != nil {
+									return *principal.UserPrincipal.PhoneNumber
```
ginaxu1: seems this logic repeated twice (once in the creation step + again when building the UserContext). Better to handle the defaulting beforehand or simplify it (in buildAuthContext)
```go
phoneNumber := ""
if principal.UserPrincipal.PhoneNumber != nil {
    phoneNumber = *principal.UserPrincipal.PhoneNumber
}
```

ginaxu1 (changes requested):

backend/internal/auth/middleware.go
sthanikan2000: Addressed

backend/internal/auth/middleware.go
```diff
@@ -73,41 +69,36 @@ func Middleware(authService *AuthService, tokenExtractor *TokenExtractor) func(h
...
-						OUHandle: principal.UserPrincipal.OUHandle,
-						OUID:     principal.UserPrincipal.OUID,
-						NSWData:  []byte(`{}`),
+					default:
+						slog.Debug("user already exists - skipping creation", "user_id", *userID)
+						authCtx.User.ID = *userID
 					}
 				}
-				authCtx.User = userCtx
 			}
```
mushrafmim, line 85:
```suggestion
if principal.UserPrincipal == nil || userProfileService == nil {
    return
}

user := principal.UserPrincipal

userID, err := userProfileService.GetOrCreateUser(
    user.UserID,
    user.Email,
    user.PhoneNumber,
    user.OUID,
)

if err != nil {
    slog.Error("failed to get or create user", "idp_user_id", user.UserID, "error", err)
    return
}

authCtx.User.ID = userID
```

Too many nested conditions, making the logic complex.

backend/internal/workflow/service/pre_consignment_service.go
```diff
@@ -301,44 +301,48 @@ func (s *PreConsignmentService) GetPreConsignmentByID(ctx context.Context, preCo
...
+			userRecord = user.Record{
+				ID:   preConsignment.TraderID,
+				Data: contextJSON,
 			}
 
-			if err := tx.Create(&uc).Error; err != nil {
-				return fmt.Errorf("failed to create user context: %w", err)
+			if err := tx.Create(&userRecord).Error; err != nil {
+				return fmt.Errorf("failed to create user record: %w", err)
 			}
```
mushrafmim: Why do we need to create a user record here?

backend/internal/workflow/service/pre_consignment_service.go
```diff
@@ -348,9 +352,9 @@ func (s *PreConsignmentService) syncTraderContextToAuth(tx *gorm.DB, preConsignm
 		return fmt.Errorf("failed to marshal updated user context: %w", err)
 	}
 
-	uc.NSWData = updatedJSON
-	if err := tx.Save(&uc).Error; err != nil {
-		return fmt.Errorf("failed to update user context: %w", err)
+	// Update the user record's NSWData field
+	if err := tx.Model(&userRecord).Update("nsw_data", updatedJSON).Error; err != nil {
+		return fmt.Errorf("failed to update user NSWData: %w", err)
```
mushrafmim: This should be removed, since we don't store company level data to a specific user anymore.

mushrafmim (changes requested):

backend/internal/profile/user/service.go
sthanikan2000: Added a UserExists check at the top of CreateUser, returning ErrUserAlreadyExists if a record is already present

backend/internal/workflow/service/pre_consignment_service.go
sthanikan2000: I remove the logics. Since we don't have company profile now

backend/internal/auth/middleware.go
sthanikan2000: Addressed

ginaxu1 (dismissed):
lgtm

backend/internal/auth/context.go
```diff
@@ -2,20 +2,63 @@ package auth
...
+	//
+	// Implementation notes:
+	//   - Should be idempotent: calling multiple times with same idpUserID should be safe
+	//   - Called during first login after token validation
+	//   - Errors are logged but don't block authentication
+	//   - Should not return error if user already exists
+	//   - If err is non-nil, the returned user ID should be nil
+	// Returns user ID of the created or existing user, or an error if the operation fails.
+	GetOrCreateUser(idpUserID, email, phone, organizationID string) (*string, error)
 }
```
mushrafmim, line 49: Shouldn't this be moved to `manager.go`?

ginaxu1 (approved):

mushrafmim (approved):

PR #461, opened 2026-05-08
[FIX][TraderApp][Frontend]: Resolve double-click requirement for CHA and HS code selection

MohamadNazik:
## Summary

This PR addresses a race condition in the search components that required users to double-click an item to select it. By utilizing `onMouseDown`, we ensure the selection is captured before the input's focus is lost, enabling reliable single-click selection.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Changes Made

- **Selection Race Condition Fix**: Migrated from `onClick` to `onMouseDown` for search results in `CHASearch.tsx` and `HSCodeSearch.tsx`. This ensures the selection event fires before the input field's `onBlur` event can unmount the results list.
- **Robust Focus Handling**: Added `onMouseDown` to search inputs to ensure consistent focus state management across different browsers (especially on Windows).

## Testing

- [x] All existing tests pass

## Checklist

## Related Issues

Closes #460 

## Additional Notes

I have reverted the initial UI/UX enhancements included in this PR to keep the focus strictly on resolving the double-click selection bug. A separate redesign of the consignment creation screen is planned, and those interface changes will be re-introduced in a separate PR.

### DIFF
--- a/portals/apps/trader-app/src/components/CHAPicker/CHASearch.tsx
+++ b/portals/apps/trader-app/src/components/CHAPicker/CHASearch.tsx
@@ -46,6 +46,7 @@ export function CHASearch({ value, onChange, options }: CHASearchProps) {
         onChange={(e) => setSearchQuery(e.target.value)}
         onFocus={() => setIsFocused(true)}
         onBlur={() => setTimeout(() => setIsFocused(false), 150)}
+        onMouseDown={() => setIsFocused(true)}
       >
         <TextField.Slot>
           <MagnifyingGlassIcon height="16" width="16" />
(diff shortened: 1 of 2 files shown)


sthanikan2000:
> > Only one CHA should be assigned to a consignment. So if one CHA is selected better to hide the `search bar` until the Selected CHA get removed. @mushrafmim what do you think?
> 
> @sthanikan2000 Instead hiding the search bar, Currently It removes the current CHA selection when a user try to type a another CHA in the search bar. Isn't that make sense that much?


Not sure @mushrafmim what do you think?

mushrafmim:
I feel we should reconsider the current multi-step approach. Right now, the flow selection is on one screen and the CHA selection is on another, even though both paths ultimately lead the user through the same process.

Multi-step forms are generally most effective when the subsequent steps differ based on the user’s earlier choices — for example, when selecting “Export” leads to a completely different set of actions than selecting “Import.” In this case, however, the overall workflow remains the same regardless of the selection.

Because of that, splitting these selections into separate steps introduces unnecessary complexity and increases cognitive load for the user without providing much functional value. Combining these related selections into a single screen could make the experience more streamlined, intuitive, and efficient.

mushrafmim (approved):
LGTM

sthanikan2000 (approved):
LGTM!

PR #480, opened 2026-05-12
feat(notifications): add notifications package with providers. 

maneesha-xyz:
## Summary

Breaking #390 into smaller PR's.

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement

## Changes Made

- List the main changes made in this PR
- Be specific about what was modified, added, or removed
- Include any architectural changes

## Testing

- [ ] I have tested this change locally
- [ ] I have tested edge cases
- [x] All existing tests pass

## Checklist

- [ ] I have checked that there are no merge conflicts

## Related Issues

#390 

## Screenshots/Demo

N/A

## Additional Notes

N/A

## Deployment Notes

N/A

### DIFF
--- a/backend/pkg/notifications/config.go
+++ b/backend/pkg/notifications/config.go
@@ -1,0 +1,14 @@
+package notifications
+
+import "fmt"
+
+type Config struct {
+	Path string
+}
+
+func (c Config) Validate() error {
+	if c.Path == "" {
+		return fmt.Errorf("NOTIFICATION_CONFIG_PATH is required")
+	}
+	return nil
+}
--- a/backend/pkg/notifications/config.go
+++ b/backend/pkg/notifications/config.go
@@ -1,14 +1,18 @@
 package notifications
 
-import "fmt"
+import "errors"
 
+var ErrConfigPathRequired = errors.New("notification config path is required")
+
+// Config holds the file-based notifications subsystem configuration.
 type Config struct {
 	Path string
 }
 
+// Validate returns ErrConfigPathRequired when Path is empty.
 func (c Config) Validate() error {
 	if c.Path == "" {
-		return fmt.Errorf("NOTIFICATION_CONFIG_PATH is required")
+		return ErrConfigPathRequired
 	}
 	return nil
 }
--- a/backend/pkg/notifications/loader.go
+++ b/backend/pkg/notifications/loader.go
@@ -6,57 +6,13 @@ import (
 	"os"
 )
 
-// expandEnv expands ${VAR_NAME} placeholders in data with their environment variable values.
-// Values are JSON-escaped before substitution so special characters cannot corrupt the JSON document.
-// Returns an error listing all unset variables if any are missing.
-func expandEnv(data []byte) ([]byte, error) {
-	var missing []string
-	expanded := os.Expand(string(data), func(key string) string {
-		val, ok := os.LookupEnv(key)
-		if !ok {
-			missing = append(missing, key)
-			return "${" + key + "}"
-		}
-		// JSON-escape the value; the template wraps ${VAR} in quotes so we strip them here.
-		// json.Marshal on a string never returns an error.
-		encoded, _ := json.Marshal(val)
-		return string(encoded[1 : len(encoded)-1])
-	})
-	if len(missing) > 0 {
-		return nil, fmt.Errorf("unset environment variables in notification config: %v", missing)
-	}
-	return []byte(expanded), nil
-}
-
-// loadConfigMap reads a JSON configuration file from the specified path, expands any
-// environment variable placeholders, and unmarshals the result into a map.
-//
-// The function performs three main steps:
-//  1. Reads the entire file contents
-//  2. Expands all ${VAR_NAME} environment variable placeholders
-//  3. Parses the resulting JSON into a map with string keys and raw JSON values
-//
-// This allows configuration files to reference environment variables which are
-// substituted at runtime, enabling flexible deployment configurations.
-//
-// Parameters:
-//   - path: The file system path to the notification configuration file
-//
-// Returns:
-//   - A map where keys are provider channel names (strings) and values are raw JSON messages
-//   - An error if the file cannot be read, environment variables cannot be expanded,
-//     or the JSON is malformed
 func loadConfigMap(path string) (map[string]json.RawMessage, error) {
 	raw, err := os.ReadFile(path)
 	if err != nil {
 		return nil, fmt.Errorf("read notification config %q: %w", path, err)
 	}
-	expanded, err := expandEnv(raw)
-	if err != nil {
-		return nil, err
-	}
 	var cfgMap map[string]json.RawMessage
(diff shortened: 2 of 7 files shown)


backend/pkg/notifications/loader.go
```diff
@@ -0,0 +1,63 @@
...
+	"os"
+)
+
+// expandEnv expands ${VAR_NAME} placeholders in data with their environment variable values.
+// Values are JSON-escaped before substitution so special characters cannot corrupt the JSON document.
+// Returns an error listing all unset variables if any are missing.
+func expandEnv(data []byte) ([]byte, error) {
+	var missing []string
+	expanded := os.Expand(string(data), func(key string) string {
+		val, ok := os.LookupEnv(key)
```
ginaxu1: antipattern to do string substitution on raw JSON before unmarshaling

better Go practice is to unmarshal the JSON into a Go struct or `map[string]any` first, and then do the `os.ExpandEnv`; or you can use existing libraries that handles env variable overriding natively and safely

ginaxu1 (changes requested):

backend/pkg/notifications/types.go
```diff
@@ -0,0 +1,36 @@
...
+
+type Request struct {
+	Channel  ChannelType
+	To       string
+	Subject  string
+	Body     string
+	HTMLBody string
+}
+
+func (r Request) Validate() error {
```
ginaxu1: you may want to use pointer receiver `func (r *Request) Validate() error {`, because a value receiver means the entire struct is copied every time Validate() is called

backend/pkg/notifications/loader.go
```diff
@@ -0,0 +1,63 @@
...
+		if !ok {
```
ginaxu1: better check if missing variable was already added, otherwise you get duplicate missing vars 
```
if !ok {
			if !slices.Contains(missing, key) {
				missing = append(missing, key)
			}
			return "${" + key + "}"
		}
```

ginaxu1 (changes requested):

backend/pkg/notifications/types.go
```diff
@@ -0,0 +1,36 @@
...
+	switch r.Channel {
+	case ChannelSMS, ChannelEmail:
+		// valid
+	default:
+		return fmt.Errorf("invalid or missing channel: %q", r.Channel)
```
ginaxu1: lines 27, 30, 33 

antipattern to use `fmt.Errorf` for static error strings, because it prevents callers from using errors.Is() to check for specific validation failures.

better do like this:
```go
var (
	ErrInvalidChannel = errors.New("invalid or missing channel")
	ErrToRequired     = errors.New("to address is required")
	ErrBodyRequired   = errors.New("body or html_body is required")
)

func (r *Request) Validate() error {
	switch r.Channel {
	case ChannelSMS, ChannelEmail:
		// valid
	default:
		return fmt.Errorf("%w: %q", ErrInvalidChannel, r.Channel)
	}
	// ...
}
```

ginaxu1 (changes requested):

backend/pkg/notifications/config.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "fmt"
+
+type Config struct {
+	Path string
+}
+
+func (c Config) Validate() error {
+	if c.Path == "" {
+		return fmt.Errorf("NOTIFICATION_CONFIG_PATH is required")
```
ginaxu1: no need to hardcode `NOTIFICATION_CONFIG_PATH`.

`Config` struct doesn't need to know how the field's populated. 

better return a generic error, using `errors.New` instead of fmt.Errorf when there is no string formatting happening: `return errors.New("notification config path is required")`

ginaxu1 (changes requested):

backend/pkg/notifications/loader.go
maneesha-xyz: Since expansion only happens inside quoted JSON strings and escaping is handled safely, a recursive unmarshaller feels like unnecessary complexity for the current use case.

backend/pkg/notifications/types.go
```diff
@@ -0,0 +1,42 @@
...
+	ChannelEmail ChannelType = "email"
+)
+
+var (
+	ErrInvalidChannel = errors.New("invalid or missing channel")
+	ErrToRequired     = errors.New("to address is required")
+	ErrBodyRequired   = errors.New("body or html_body is required")
+)
+
+type Request struct {
```
ginaxu1, line 21: not convinced this is best approach: current `Request` struct forces all channels to share the exact same payload format. SMS doesn't need have Subject or an HTMLBody

@maneeshaxyz what do you think about
```
type EmailPayload struct {
	To       string
	Subject  string
	Body     string
	HTMLBody string
}

type SMSPayload struct {
	To   string
	Body string
}

// Update the interface to accept a generic payload or specific methods
type Provider interface {
	Type() ChannelType
	Configure(cfg json.RawMessage) error
	Send(ctx context.Context, payload any) error // implementation type-asserts the payload
}
```

ginaxu1 (changes requested):

backend/pkg/notifications/types.go
ginaxu1: always make sure you add struct tags (because relying on Go's implicit field matching is brittle)

```
Channel  ChannelType `json:"channel"`
To       string      `json:"to"`
Subject  string      `json:"subject,omitempty"`
Body     string      `json:"body"`
HTMLBody string      `json:"html_body,omitempty"`
```

backend/pkg/notifications/loader.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+)
+
+// expandEnv expands ${VAR_NAME} placeholders in data with their environment variable values.
+// Values are JSON-escaped before substitution so special characters cannot corrupt the JSON document.
+// Returns an error listing all unset variables if any are missing.
+func expandEnv(data []byte) ([]byte, error) {
```
ginaxu1: on 2nd thought why can't we use a standard configuration library like viper or envconfig to safely handle this?

ginaxu1 (changes requested):

sthanikan2000 (dismissed):
lgtm

backend/pkg/notifications/types.go
sthanikan2000: @ginaxu1 Let's not worry about this now. 
`type.go` should have Request struct only and it should be like follows:
```go
type Request struct {
    TemplateCode string
	UserID string
	Channels []ChannelType
	Data map[string]any
}
```

CC: @mushrafmim , @maneeshaxyz

sthanikan2000 (approved):
lgtm

ginaxu1 (approved):

PR #493, opened 2026-05-13
feat(oga): serve OGA frontend from backend with optional HEADLESS build

### LINKED ISSUE #40: Serve Agency Frontend static assets from the backend binary
### Current Limitation

The OGA app frontend (`portals/apps/oga-app`) is built independently via Vite and must be served by a separate process, such as nginx, a CDN, or a local dev server. The backend (`oga/cmd/server`) only registers `/api/oga/...` routes and has no mechanism to serve static frontend assets. This creates operational overhead — even in development and simple deployments, two separate servers must be running and coordinated. There is also no SPA fallback (serving `index.html` for client-side routes), which means deep-linking and page refreshes break unless a separate server is configured to handle them.

### Suggested Improvement

Embed the OGA app's compiled output (`dist/`) into the backend binary using Go's `//go:embed` directive and serve it via `http.FileServer` with an SPA fallback.

### Concretely
- Add a `backend/internal/static/` package or a dedicated `embed.go` file that uses `//go:embed` to embed the OGA app build output.
- Register a catch-all route in `cmd/server/main.go` after all `/api/oga/...` routes that serves the embedded filesystem.
- Implement an SPA fallback handler so that any path not matching a real file is served `index.html`, preserving client-side routing.
- Add a build step in `Makefile` or CI that runs `vite build` inside `portals/apps/oga-app` before `go build`, so the `dist/` directory is always present when the binary is compiled.
- Optionally gate the static serving behind a config flag, such as `OGA_SERVE_STATIC=true`, to keep the API-only mode available for deployments that prefer a dedicated CDN.

### Version

_No response_

### Additional Context

- The backend entry point is `oga/cmd/server/main.go`. Routes are registered on a plain `http.ServeMux`.
- The OGA app is a Vite + React SPA, and its build output is `portals/apps/oga-app/dist/`.
- Serving both API and frontend from a single binary simplifies Docker images and local development (`go run ./cmd/server` is enough).

saknarajapakshe:
## Summary

This PR enables the OGA backend to serve the OGA frontend (SPA) as static files, while keeping the Docker image flexible via a `HEADLESS` build argument. When `HEADLESS=true` (default), the image skips the frontend build — useful for API-only deployments and CI. When `HEADLESS=false`, the Dockerfile compiles the OGA frontend via pnpm and embeds the output, which the Go server then serves directly.

## Type of Change

## Changes Made

- **`oga/Dockerfile`**: Introduced multi-stage build with a `HEADLESS` ARG (default `true`):
  - `assets-false` stage: installs Node 22, runs `pnpm install` + `pnpm run build` for the OGA app under `portals/`
  - `assets-true` stage: creates an empty `dist/` directory as a no-op placeholder
  - `final-assets` selector stage chooses between the two based on the ARG
  - Final image copies the selected assets into `/app/public`
  - Fixed `go.mod`/`go.sum` and `data/forms` paths to be relative to `oga/` (monorepo layout)
- **`oga/cmd/server/main.go`**: Added a catch-all route that serves the `public/` directory:
  - Requests to `/api/` or paths with a file extension are served directly by `http.FileServer`
  - All other paths fall back to `public/index.html` to support SPA client-side routing

## Testing

- [x] All existing tests pass

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas

## Related Issues

Closes OpenNSW/nsw-agency#40

## Additional Notes

- `HEADLESS=true` is the safe default — existing deployments that do not need the frontend are unaffected and do not pay the npm build cost.
- The SPA fallback uses `strings.Contains(r.URL.Path, ".")` as a heuristic for static asset requests.

## Deployment Notes

To build the image with the frontend bundled in:
```
docker build --build-arg HEADLESS=false -t oga:latest .
```
Default (API-only, no frontend assets):
```
docker build -t oga:latest .
```

### DIFF
--- a/oga/cmd/server/main.go
+++ b/oga/cmd/server/main.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
@@ -98,6 +99,19 @@ func main() {
 
 	mux.HandleFunc("POST /api/oga/uploads", storageHandler.HandleCreateUpload)
 	mux.HandleFunc("GET /api/oga/uploads/{key}", storageHandler.HandleGetUploadURL)
+
+	// Serve static files from public/ folder (build output of frontend)
+	fs := http.FileServer(http.Dir("public"))
+	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// If the request is for an API or appears to be a static asset with an extension,
+		// let the FileServer handle it.
+		if strings.HasPrefix(r.URL.Path, "/api/") || strings.Contains(r.URL.Path, ".") {
+			fs.ServeHTTP(w, r)
+			return
+		}
+		// Otherwise, fallback to index.html to support SPA client-side routing.
+		http.ServeFile(w, r, "public/index.html")
+	}))
 
 	// Set up graceful shutdown
 	serverAddr := fmt.Sprintf(":%s", cfg.Port)
--- a/oga/Dockerfile
+++ b/oga/Dockerfile
@@ -1,3 +1,22 @@
+ARG HEADLESS=true
+
+# --- Stage: Build OGA frontend (active when HEADLESS=false) ---
+FROM node:22-bookworm-slim AS assets-false
+WORKDIR /app
+RUN npm install -g pnpm
+COPY portals/ portals/
+RUN cd portals && CI=true pnpm install --frozen-lockfile
+RUN cd portals/apps/oga-app && pnpm run build
+# dist/ is now at: portals/apps/oga-app/dist/
+
+# --- Stage: Empty assets (active when HEADLESS=true) ---
+FROM alpine:latest AS assets-true
+RUN mkdir -p /app/portals/apps/oga-app/dist
+
+# --- Asset selector ---
+FROM assets-${HEADLESS} AS final-assets
+
+# --- Stage: Build Go binary ---
 FROM golang:1.25-bookworm AS builder
 
 WORKDIR /app
@@ -6,8 +25,10 @@ COPY go.mod go.sum ./
 RUN go mod download
 
 COPY . .
+
 RUN go build -o /out/oga ./cmd/server
 
+# --- Final image ---
 FROM debian:bookworm-slim
 
 LABEL org.opencontainers.image.source="https://github.com/OpenNSW/nsw"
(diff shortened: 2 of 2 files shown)


oga/cmd/server/main.go
```diff
@@ -99,6 +100,19 @@ func main() {
 	mux.HandleFunc("POST /api/oga/uploads", storageHandler.HandleCreateUpload)
 	mux.HandleFunc("GET /api/oga/uploads/{key}", storageHandler.HandleGetUploadURL)
 
+	// Serve static files from public/ folder (build output of frontend)
+	fs := http.FileServer(http.Dir("public"))
+	mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// If the request is for an API or appears to be a static asset with an extension,
+		// let the FileServer handle it.
+		if strings.HasPrefix(r.URL.Path, "/api/") || strings.Contains(r.URL.Path, ".") {
```
ginaxu1, line 108: antipattern to check strings for SPA fallback, better to do `os.Stat` to check if the requested file actually exists on disk. If it does, let http.FileServer handle it. If it doesn't, fall back to index.html

one way to re-write lines 91 to 100: a custom FileSystem handler
```
mux.Handle("/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	// Return 404 for unmatched /api/ routes immediately
	if strings.HasPrefix(r.URL.Path, "/api/") {
		http.NotFound(w, r)
		return
	}

	path := filepath.Join("public", filepath.Clean(r.URL.Path))
	
	// Check if file exists on disk
	info, err := os.Stat(targetPath)
	if os.IsNotExist(err) || info.IsDir() {
		// Handle HEADLESS=true scenario: if index.html doesn't exist, act as a pure API
		if _, indexErr := os.Stat(indexPath); os.IsNotExist(indexErr) {
		  http.NotFound(w, r)
		  return
		}

		// SPA Fallback: Serve index.html with strict NO CACHE headers
		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
		http.ServeFile(w, r, indexPath)
		return
	} else if err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}

	// File exists, serve it natively
	http.FileServer(http.Dir("public")).ServeHTTP(w, r)
}))
```

ginaxu1 (changes requested):

PR #515, opened 2026-05-18
Add publish-helm.yml to publish Helm Charts to GHCR

ginaxu1:
## Summary
Add `.github/workflows/publish-helm.yml` workflow to automate the packaging and publishing of our Helm subcharts to GHCR secure, remote OCI artifacts (`oci://ghcr.io/opennsw/charts`) whenever changes are merged into the `main` branch under the `deployments/helm/**` directory

## Why we want this

Currently, the `nsw-gitops` umbrella chart [envs/base/apps-umbrella/Chart.yaml](https://github.com/OpenNSW/nsw-gitops/blob/main/envs/base/apps-umbrella/Chart.yaml#L10) relies on locally bundled `.tgz` files** stored under the `charts/` directory. This causes issues including the following
1. any modifications to the Helm charts require manual repackaging via `helm package` and manual commits/pushes of binarily encoded `.tgz` files directly to the `nsw-gitops` repository. This is highly error-prone and slow
2. storing binary `.tgz` packages directly inside Git causes massive, unnecessary repository bloat and pollutes git history
4. Code changes to app manifests are coupled directly to the GitOps sync configurations, not following GitOps best practices where application definitions and env configurations should be cleanly separated

## Changes
By moving to an **OCI-based publishing model** via `publish-helm.yml`, we can:
* publish every chart change directly to GHCR as a versioned artifact automatically upon merging to `main`
* have the `nsw-gitops` repository to point directly to these remote OCI packages, completely eliminating the need for checked-in `.tgz` files
* allow Renovate bot to automatically detect newer chart versions in GHCR and automatically open PRs to bump charts in `nsw-gitops`, achieving fully **automated GitOps deployments for Helm chart changes**

## Verification
1. verify the workflow is successfully triggered upon merging this branch to `main`.
2. confirm that the charts (`oga-app`, `oga-backend`, `trader-app`, etc.) appear under the GHCR Packages page (`ghcr.io/opennsw/charts/`)
3. ArgoCD Reference Update: once published, update the dependencies in `envs/base/apps-umbrella/Chart.yaml` to reference:
   ```yaml
   dependencies:
     - name: trader-app
       version: "0.1.0"
       repository: "oci://ghcr.io/opennsw/charts"
   ```
   and run a test sync in `nsw-dev` to verify the remote fetching works.

### DIFF
--- a/.github/workflows/publish-helm.yml
+++ b/.github/workflows/publish-helm.yml
@@ -1,0 +1,55 @@
+name: Publish Helm Charts
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'deployments/helm/**'
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  publish:
+    name: Publish Helm Charts to GHCR
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: v3.14.0
+
+      - name: Convert repository owner to lowercase
+        id: repo_owner
+        run: echo "owner=$(echo '${{ github.repository_owner }}' | tr '[:upper:]' '[:lower:]')" >> "$GITHUB_OUTPUT"
+
+      - name: Log in to GHCR
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ${{ env.REGISTRY }} -u ${{ github.actor }} --password-stdin
+
+      - name: Package and Publish Charts
+        run: |
+          # Iterate through each directory in deployments/helm
+          for chart_dir in deployments/helm/*; do
+            if [ -d "$chart_dir" ]; then
+              chart_name=$(basename "$chart_dir")
+              echo "Packaging chart: $chart_name"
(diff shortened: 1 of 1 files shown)


sthanikan2000 (changes requested):
Can you go through the following issue point out by the `claude`, when I review with it?

**1. No dependency-aware triggering**
The workflow only watches `deployments/helm/**`, so changes to application code that affect a chart (e.g. a new image tag) won't trigger a republish. Chart updates and app updates need to be kept in sync manually.

**2. No version bump validation**
There's no check that the chart version in `Chart.yaml` has actually been incremented before publishing. Pushing the same version will either silently overwrite or get rejected by GHCR, depending on registry settings.

**3. No chart validation before publishing**
There's no `helm lint` or `helm template` step. A malformed chart will be packaged and pushed without any warning.

**4. Silent failures in the packaging loop**
If `helm package` fails for one chart, the script continues to the next due to no error handling. A broken chart could be silently skipped while the workflow reports success.

**5. No verification after push**
There's no confirmation step (e.g. `helm pull`) to verify the chart was actually published and is pullable from GHCR after the push.

**6. Packaged charts directory not cleaned up**
`./packaged-charts/` is never explicitly created with error handling or cleaned up. If a previous run's artifacts somehow persist, stale `.tgz` files could be pushed unintentionally.

PR #531, opened 2026-05-19
feat(templatesource): introduce pkg/templatesource with local and GitHub-backed sources

sthanikan2000:
## Summary

Introduces `backend/pkg/templatesource` — a small, self-contained package that resolves JSON template blobs by ID. It is intentionally free of OGA-specific config or env coupling so it can be reused by other services.

Two implementations are provided:
- **`NewLocal`** — reads `.json` files from a directory into memory at startup (development / offline use).
- **`NewGitHub`** — fetches `manifest.json` from a GitHub repository at startup (fail-fast), then lazily fetches and caches individual template files on demand with optional background manifest refresh.

## Type of Change

## Changes Made

- `source.go` — defines the `Source` interface with a documented three-state return contract (`found / not-found / error`).
- `local.go` — `NewLocal` implementation: reads all `.json` files from a directory, validates JSON at load time, skips subdirectories and non-JSON files.
- `github.go` — `NewGitHub` implementation: manifest-driven loader with lazy per-template fetching, in-memory cache keyed by repo-relative path (so manifest refreshes automatically invalidate stale paths), `sync.RWMutex` for concurrent safety, `io.LimitReader` cap (10 MiB) to guard against runaway responses, and idempotent `Close` via `sync.Once`.
- `local_test.go` / `github_test.go` — full test suite using `net/http/httptest`; no real network calls required.
- `README.md` — usage examples, `GitHubConfig` field reference, and `Source` contract documentation.

## Testing

- [x] All existing tests pass

**Coverage: 98.2%** (`go test ./pkg/templatesource/... -covermode=atomic`)

Key scenarios covered:
- Local: valid load, raw JSON round-trip, non-JSON file skipping, subdirectory exclusion, missing dir, invalid JSON, unreadable file, `Close`.
- GitHub: happy path + cache hit, unknown ID, missing manifest (fail-fast), invalid manifest JSON (fail-fast), missing `byId` field, template fetch error, invalid template JSON, manifest refresh cache invalidation, background refresh ticking, background refresh error logging, invalid `BaseURL`, default `BaseURL` fallback, concurrent close idempotency.

## Checklist

## Related Issues

Related to #513 

## Additional Notes

The two remaining uncovered statements (≈1.8%) are the concurrency double-check cache path in `GetTemplate` (requires a precise goroutine race to trigger deterministically) and the `http.NewRequestWithContext` failure path (unreachable in normal operation since URLs are constructed via validated `url.JoinPath`).

### DIFF
--- a/backend/pkg/templatesource/README.md
+++ b/backend/pkg/templatesource/README.md
@@ -33,7 +33,7 @@ if err != nil {
 }
 defer src.Close()
 
-raw, ok, err := src.GetTemplate(ctx, "build-licence")
+raw, ok, err := src.GetTemplate(context.Background(), "build-licence")
 ```
 
 - Returns an error and refuses to start if `dir` is missing or any file
@@ -48,7 +48,7 @@ raw, ok, err := src.GetTemplate(ctx, "build-licence")
 first access.
 
 ```go
-src, err := templatesource.NewGitHub(ctx, templatesource.GitHubConfig{
+src, err := templatesource.NewGitHub(context.Background(), templatesource.GitHubConfig{
     Repo:            "OpenNSW/one-trade-templates",
     Ref:             "abc1234",           // pin to a SHA in production
     RefreshInterval: 5 * time.Minute,    // 0 disables background refresh
@@ -57,6 +57,8 @@ if err != nil {
     log.Fatal(err)
 }
 defer src.Close()
+
+raw, ok, err := src.GetTemplate(context.Background(), "build-licence")
 ```
 
 #### `GitHubConfig` fields
--- a/backend/pkg/templatesource/README.md
+++ b/backend/pkg/templatesource/README.md
@@ -8,10 +8,8 @@ interface with two implementations:
 | Local filesystem | `NewLocal(dir)` | Development / offline testing |
 | GitHub raw content | `NewGitHub(ctx, cfg)` | Staging and production |
 
-The package returns opaque `json.RawMessage` bytes and never inspects the JSON
-shape, so it can serve forms, workflow definitions, or any other
-manifest-keyed artifact. It is intentionally free of OGA-specific config or
-env coupling so it can be reused by other services.
+Each implementation validates that blobs are syntactically valid JSON but never
+inspects the document structure, leaving interpretation entirely to the caller.
 
 ## Import
 
--- a/backend/pkg/templatesource/README.md
+++ b/backend/pkg/templatesource/README.md
@@ -1,0 +1,101 @@
+# templatesource
+
+Resolves JSON template blobs by ID. The package exposes a single `Source`
+interface with two implementations:
+
+| Implementation | Constructor | When to use |
+|---|---|---|
+| Local filesystem | `NewLocal(dir)` | Development / offline testing |
+| GitHub raw content | `NewGitHub(ctx, cfg)` | Staging and production |
+
+The package returns opaque `json.RawMessage` bytes and never inspects the JSON
+shape, so it can serve forms, workflow definitions, or any other
+manifest-keyed artifact. It is intentionally free of OGA-specific config or
+env coupling so it can be reused by other services.
+
+## Import
+
+```go
+import "github.com/OpenNSW/nsw/pkg/templatesource"
+```
+
+## Usage
+
+### Local source
+
+`NewLocal` reads every `.json` file in `dir` into memory at startup. The file
+basename without the `.json` extension becomes the template ID.
+
+```go
+src, err := templatesource.NewLocal("/etc/oga/templates")
+if err != nil {
+    log.Fatal(err)
+}
+defer src.Close()
+
+raw, ok, err := src.GetTemplate(ctx, "build-licence")
+```
+
+- Returns an error and refuses to start if `dir` is missing or any file
+  contains invalid JSON.
+- Subdirectories and non-`.json` files are silently skipped.
+- `Close` is a no-op but must still be called to satisfy the interface.
+
+### GitHub source
+
+`NewGitHub` fetches `manifest.json` from a GitHub repository at startup
+(fail-fast), then lazily fetches and caches individual template files on
+first access.
+
+```go
+src, err := templatesource.NewGitHub(ctx, templatesource.GitHubConfig{
+    Repo:            "OpenNSW/one-trade-templates",
+    Ref:             "abc1234",           // pin to a SHA in production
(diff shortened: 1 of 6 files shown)


backend/pkg/templatesource/local.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, entry := range entries {
+		if entry.IsDir() || !strings.HasSuffix(entry.Name(), ".json") {
+			continue
+		}
+
+		data, err := os.ReadFile(filepath.Join(dir, entry.Name()))
+		if err != nil {
+			return nil, fmt.Errorf("failed to read template file %q: %w", entry.Name(), err)
+		}
+		if !json.Valid(data) {
```
ginaxu1, line 37: nit: inconsistency in README.me where you wrote _"The package returns opaque `json.RawMessage` bytes and never inspects the JSON shape,"_ and implementation here


double check the approach - is it better to validate JSON here at load time?? Or remove such validation, read the file bytes into memory instead, then let the caller handle the JSON parsing errors?

ginaxu1 (changes requested):

backend/pkg/templatesource/github.go
```diff
@@ -0,0 +1,229 @@
...
+		case <-s.done:
+			return
+		case <-ticker.C:
+			s.refresh()
+		}
+	}
+}
+
+func (s *githubSource) refresh() {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
ginaxu1: better to derive the context from the `githubSource` lifecycle. Add a context.Context and cancelFunc to the `githubSource` struct that is cancelled in Close().

```// in NewGitHub
src.ctx, src.cancel = context.WithCancel(context.Background())

// in refresh():
ctx, cancel := context.WithTimeout(s.ctx, 10*time.Second)
```

ginaxu1 (changes requested):

backend/pkg/templatesource/local.go
```diff
@@ -0,0 +1,58 @@
...
+// a Source that serves them. The file basename (without ".json") is the
+// template ID. Returns an error if dir is missing or any file contains
+// invalid JSON.
+func NewLocal(dir string) (Source, error) {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read templates directory %q: %w", dir, err)
+	}
+
+	templates := make(map[string]json.RawMessage)
```
ginaxu1, line 27: should check if directory contains at least 1 template 
```
if len(templates) == 0 {
    return nil, fmt.Errorf("templatesource: no .json files found in %q", dir)
}
```

ginaxu1 (changes requested):

backend/pkg/templatesource/github.go
```diff
@@ -0,0 +1,234 @@
...
+	// still appears in byId, its bytes are still correct.
+	newPaths := make(map[string]struct{}, len(m.ByID))
+	for _, p := range m.ByID {
+		newPaths[p] = struct{}{}
+	}
+	for path := range s.templateCache {
+		if _, ok := newPaths[path]; !ok {
+			delete(s.templateCache, path)
+		}
+	}
```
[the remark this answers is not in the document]
sthanikan2000: Fixed in bc5c1f47 — replaced the selective path-based eviction with a full `templateCache = make(...)` reset on every manifest refresh. Added `TestGitHub_ManifestRefreshClearsStaleContent` which reproduces the exact scenario (same path, updated bytes) and would have failed against the old logic.

backend/pkg/templatesource/README.md
```diff
@@ -0,0 +1,101 @@
...
+basename without the `.json` extension becomes the template ID.
+
+```go
+src, err := templatesource.NewLocal("/etc/oga/templates")
+if err != nil {
+    log.Fatal(err)
+}
+defer src.Close()
+
+raw, ok, err := src.GetTemplate(ctx, "build-licence")
```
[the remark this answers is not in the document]
sthanikan2000: Fixed — replaced `ctx` with `context.Background()` in the local source code snippet.

backend/pkg/templatesource/README.md
```diff
@@ -0,0 +1,101 @@
...
+```go
+src, err := templatesource.NewGitHub(ctx, templatesource.GitHubConfig{
+    Repo:            "OpenNSW/one-trade-templates",
+    Ref:             "abc1234",           // pin to a SHA in production
+    RefreshInterval: 5 * time.Minute,    // 0 disables background refresh
+})
+if err != nil {
+    log.Fatal(err)
+}
+defer src.Close()
```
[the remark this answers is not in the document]
sthanikan2000: Fixed — replaced `ctx` with `context.Background()` in the GitHub constructor call and added a `GetTemplate` call after `defer src.Close()` to mirror the local source example.

backend/pkg/templatesource/local.go
sthanikan2000: Good catch. we validate JSON *syntax* (so callers never receive a malformed blob), but we never inspect the document *structure* (fields, types, nesting). Updated the package doc comment in `source.go` and the README to say exactly that — removed the OGA/domain-specific wording too to keep the package fully generic.

backend/pkg/templatesource/github.go
sthanikan2000: Done in f99d213c. Went a step further than the suggestion: replaced `done chan struct{}` and `closeOnce sync.Once` entirely with `ctx`/`cancel` on the struct. `Close()` now calls `s.cancel()` (idempotent, so `closeOnce` is no longer needed), `refreshLoop` selects on `s.ctx.Done()`, and `refresh()` derives its 10 s timeout from `s.ctx` so any in-flight fetch is cancelled on shutdown.

backend/pkg/templatesource/local.go
sthanikan2000: Added in f99d213c. `NewLocal` now returns `fmt.Errorf("templatesource: no .json files found in %q", dir)` when the template map is empty after the loop, covering both a genuinely empty directory and a directory that contains only non-JSON files. Two new tests (`TestLocal_ErrorOnEmptyDir`, `TestLocal_ErrorOnDirWithNoJSONFiles`) cover both cases.

mushrafmim (approved):

ginaxu1 (approved):

PR #539, opened 2026-05-21
refactor(consignment): scope consignments by trader/CHA company

### LINKED ISSUE #331: Improvement: Persist GlobalContext in Consignment Shell and add CHA validation
### Current Limitation

Currently, the \`Consignment\` model does not store the \`GlobalContext\` when a Trader creates a consignment shell (Stage 1). Instead, it is inconsistently passed later during initialization (Stage 2) and only sent downstream to the workflow managers. 

Furthermore, during the Stage 2 initialization (\`InitializeConsignmentByID\`), there is a lacking authorization check to verify that the \`CHAID\` assigned to the consignment actually matches the CHA user trying to initialize it.

### Suggested Improvement
1. Add a \`GlobalContext map[string]any\` field to the \`model.Consignment\` struct with JSONB storage.
2. Update \`CreateConsignmentShell\` (Router & Service) to accept the trader's \`GlobalContext\` from \`authCtx\` and persist it with the shell.
3. Update \`InitializeConsignmentByID\` (Router & Service) to resolve the requesting user's CHA ID (via their email).
4. Add authorization logic to ensure \`Consignment.CHAID == requestingChaID\`, returning a Forbidden error on mismatch.
5. Reuse the shell's persisted \`GlobalContext\` to start the workflow instead of rebuilding it in Stage 2.

### Version

_No response_

### Additional Context

This addresses a known TODO in \`consignment_router.go\` inside \`HandleInitializeConsignment\` and standardizes how we manage trader context across the two-stage consignment creation process."

sthanikan2000:
## Summary

Builds on the profile foundation from #533 by giving every consignment both a trader-company and a CHA-company identity. Stage 1 (CreateConsignmentShell) records who-belongs-where; Stage 2 (InitializeConsignmentByID) verifies the picking CHA belongs to the chosen CHA company before assigning. Listing semantics become company-scoped so colleagues in the same trader company see each other's consignments, and any CHA in the same CHA company sees all assigned ones.

This is the second of three PRs in the consignment ↔ company wiring stack. The remaining PR (trader-app frontend — switches the CHA picker to `GET /api/v1/companies?has_cha=true&name=...`, renames the create payload field `chaId` → `chaCompanyId`) is queued locally and will go up once this merges.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

Breaking notes:
- `consignments` table schema changes (new NOT NULL FKs); requires a schema rebuild in dev environments.
- `CreateConsignmentDTO.chaId` is renamed to `chaCompanyId`. Any client sending the old field will hit the new "chaCompanyId is required" validation. The trader-app fix is the next (and final) PR in this stack.
- `InitializeConsignmentByID` signature drops the unused `globalContext` parameter and adds `chaID`; existing call sites need updating (the router is updated in this PR; no other callers).
- `Filter.TraderID` / `Filter.ChaID` are replaced with `Filter.TraderCompanyID` / `Filter.CHACompanyID`.
- `GetConsignmentsByTraderID` is removed — no remaining callers.

## Changes Made

- **Schema** ([`001_initial_schema.up.sql`](backend/internal/database/migrations/001_initial_schema.up.sql)) — `consignments` gains `trader_company_id` and `cha_company_id` (both NOT NULL FKs to `company_records(id)`); `cha_id` drops NOT NULL (no longer set at Stage 1); matching indexes added.
- **Model** ([`consignment/model.go`](backend/internal/consignment/model.go)) — `Consignment` struct reordered, gains `TraderCompanyID` / `CHACompanyID`; the embedded `cha.Record` relation is dropped (no callers). `CreateConsignmentDTO` renames `ChaID` → `ChaCompanyID`. `Filter` replaces `TraderID` / `ChaID` with `TraderCompanyID` / `CHACompanyID`. `DetailDTO` / `SummaryDTO` surface the new company IDs.
- **Errors** ([`consignment/errors.go`](backend/internal/consignment/errors.go) — new) — sentinels `ErrCompanyNotCHA`, `ErrCHACompanyMismatch`.
- **Service** ([`consignment/service.go`](backend/internal/consignment/service.go)) — `NewService` now takes `company.Service` and `user.Service`.
  - `CreateConsignmentShell(ctx, flow, chaCompanyID, traderID)`: looks up the CHA company, rejects when `has_cha = false`, resolves the trader's company from their `ou_handle`, persists both company IDs.
  - `InitializeConsignmentByID(ctx, consignmentID, hsCodeIDs, chaID)`: drops `globalContext`; verifies CHA via `GetByID` and `cha.CompanyID == consignment.CHACompanyID`; loads trader company and passes `{"traderCompany": <marshalled record>}` to `StartWorkflow`.
  - `ListConsignments` filters by `trader_company_id` / `cha_company_id`.
- **Router** ([`consignment/router.go`](backend/internal/consignment/router.go)) — gains `company.Service`. `HandleCreateConsignment` maps `company.ErrCompanyNotFound` → 404 and `ErrCompanyNotCHA` → 400. `HandleInitializeConsignment` resolves the CHA via `cha.GetByEmail(authCtx.Email)` and passes the resolved ID; `ErrCHACompanyMismatch` → 403. `HandleGetConsignments` resolves the requesting user's company by `ou_handle` and scopes both roles by company.
- **Bootstrap** ([`bootstrap/app.go`](backend/internal/app/bootstrap/app.go)) — wires `companyService` + `userProfileService` into `consignment.NewService` and `consignment.NewRouter`.

## Testing

- [x] All existing tests pass

Verification performed:
- `cd backend && go build ./... && gofmt -l . && go vet ./... && go test ./... && golangci-lint run ./...` — all green.
- New tests cover: `CreateConsignmentShell` happy + `CompanyNotFound` + `CompanyNotCHA` + insert error; `InitializeConsignmentByID` happy + `NoTemplate` + `StartWorkflowError` + `TemplateProviderError` + new `CHACompanyMismatch`; company-scoped `ListConsignments`; `HandleGetConsignments` resolving company by `OUHandle` for both roles; `HandleCreateConsignment` propagating `company.ErrCompanyNotFound` → 404.
- End-to-end smoke against a fresh DB seeded by `idp/02-sample-resources.sh`:
  - As Suresh (trader at ADAM PVT LTD): `POST /api/v1/consignments {flow:"IMPORT", chaCompanyId:"adam-pvt-ltd"}` succeeds; row has `trader_company_id = adam-pvt-ltd`, `cha_company_id = adam-pvt-ltd`, `cha_id IS NULL`.
  - As Ramesh (CHA at ADAM PVT LTD): `PUT /api/v1/consignments/{id}` succeeds and stamps `cha_id`.
  - As Naresh (CHA at EDWARD PVT LTD) attempting the same: 403 (`ErrCHACompanyMismatch`).
  - As Gomesh (trader at ADAM PVT LTD, not the creator): `GET /api/v1/consignments?role=trader` includes Suresh's consignment.

## Checklist

## Related Issues

Closes #331

## Screenshots/Demo

N/A — backend / schema change. The user-facing impact is rolled into the trader-app PR that follows.

## Additional Notes

One PR remains in this stack:
- **trader-app frontend** — switches the CHA picker to `GET /api/v1/companies?has_cha=true&name=...` (debounced server-side filter), renames the create-consignment payload field `chaId` → `chaCompanyId`, and trims the `Company` interface to match the new `Summary` response shape. Branch is already prepared locally; will be opened once this merges.

The empty-string-vs-NULL pitfall for `cha_id` on Stage 1 inserts is handled at the GORM model level — `CHAID` keeps its `string` type but the column is now nullable, and the fix you applied during E2E testing ensures the insert emits `NULL` instead of `''`.

## Deployment Notes

- Requires a **schema rebuild** in dev environments because `consignments` gains two NOT NULL FK columns (no in-place migration). For local Postgres: `backend/internal/database/migrations/down.sh` then `run.sh`. No production data is affected — the schema change is dev/sample-data only at this stage.
- Any out-of-tree HTTP client that constructs the Stage 1 payload must send `chaCompanyId` instead of `chaId`. The trader-app fix lands in the next PR; no other in-repo callers exist.

### DIFF
--- a/backend/internal/app/bootstrap/app.go
+++ b/backend/internal/app/bootstrap/app.go
@@ -85,6 +85,7 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 	templateService := service.NewTemplateService(db)
 	chaService := cha.NewService(db)
 	companyService := company.NewService(db)
+	userProfileService := user.NewService(db)
 	hsCodeService := hscode.NewService(db)
 
 	temporalClient, err := temporal.NewClient(cfg.Temporal)
@@ -93,8 +94,8 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 		return nil, fmt.Errorf("failed to create temporal client: %w", err)
 	}
 
-	consignmentService := consignment.NewService(db, templateService, chaService, hsCodeService)
-	consignmentRouter := consignment.NewRouter(consignmentService, chaService)
+	consignmentService := consignment.NewService(db, templateService, chaService, companyService, userProfileService, hsCodeService)
+	consignmentRouter := consignment.NewRouter(consignmentService, chaService, companyService)
 
 	workflowRuntime, err := workflowruntime.NewRuntime(temporalClient, tm, templateService, consignmentService)
 	if err != nil {
@@ -126,8 +127,6 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 	storageHandler := storage.NewHTTPHandler(storageService)
 
 	paymentHandler := payments.NewHTTPHandler(paymentService)
-
-	userProfileService := user.NewService(db)
 
 	authManager, err := auth.NewManager(userProfileService, cfg.Auth)
 	if err != nil {
--- a/backend/internal/consignment/model.go
+++ b/backend/internal/consignment/model.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/OpenNSW/nsw/internal/hscode"
-	"github.com/OpenNSW/nsw/internal/profile/cha"
 	"github.com/OpenNSW/nsw/internal/workflow/model"
 )
 
@@ -34,14 +33,18 @@ type Consignment struct {
 	CreatedAt time.Time `gorm:"type:timestamptz;column:created_at;not null;autoCreateTime" json:"createdAt"`
 	UpdatedAt time.Time `gorm:"type:timestamptz;column:updated_at;not null;autoUpdateTime" json:"updatedAt"`
 
-	Flow     Flow   `gorm:"type:varchar(50);column:flow;not null" json:"flow"`             // e.g., IMPORT, EXPORT
-	TraderID string `gorm:"type:varchar(100);column:trader_id;not null" json:"traderId"`   // ID of the trader associated with the consignment
-	State    State  `gorm:"type:varchar(50);column:state;not null" json:"state"`           // State of the consignment
-	Items    []Item `gorm:"type:jsonb;column:items;serializer:json;not null" json:"items"` // Items in the consignment
+	// Core attributes
+	Flow  Flow   `gorm:"type:varchar(50);column:flow;not null" json:"flow"`             // IMPORT or EXPORT
+	State State  `gorm:"type:varchar(50);column:state;not null" json:"state"`           // INITIALIZED → IN_PROGRESS → FINISHED
+	Items []Item `gorm:"type:jsonb;column:items;serializer:json;not null" json:"items"` // Items in the consignment
 
-	// CHA (Customs House Agent) – set at Stage 1 by Trader; CHA completes Stage 2 by selecting HS Code
-	CHAID string     `gorm:"type:varchar(100);column:cha_id;not null" json:"chaId"` // Assigned CHA (Stage 1)
-	CHA   cha.Record `gorm:"foreignKey:CHAID" json:"cha"`                           // Associated CHA entity
+	// Trader (set at Stage 1)
+	TraderID        string `gorm:"type:varchar(100);column:trader_id;not null" json:"traderId"`                // Trader user who created the consignment
+	TraderCompanyID string `gorm:"type:varchar(100);column:trader_company_id;not null" json:"traderCompanyId"` // Company the trader belongs to
+
+	// CHA (company chosen at Stage 1; specific CHA assigned at Stage 2)
+	CHACompanyID string  `gorm:"type:varchar(100);column:cha_company_id;not null" json:"chaCompanyId"` // CHA company selected by the trader at Stage 1
+	CHAID        *string `gorm:"type:varchar(100);column:cha_id" json:"chaId,omitempty"`               // CHA who claimed the consignment at Stage 2
 
 	// Relationships
 	Workflow *model.Workflow `gorm:"foreignKey:ID;references:ID" json:"-"` // Associated Workflow (1:1, same ID)
@@ -72,17 +75,17 @@ type CreateConsignmentItemDTO struct {
 }
 
 // CreateConsignmentDTO represents the data required to create a consignment.
-// Stage 1 (two-stage flow): provide flow + chaId only → creates shell with state INITIALIZED.
-// Legacy / single-stage: provide flow + items → creates consignment and initializes workflow.
+// Stage 1 (two-stage flow): provide flow + chaCompanyId → creates shell with state INITIALIZED.
+// A specific CHA is assigned later, at Stage 2, when a CHA from chaCompanyId claims the consignment.
 type CreateConsignmentDTO struct {
-	Flow  Flow                       `json:"flow"`
-	ChaID string                     `json:"chaId"`
-	Items []CreateConsignmentItemDTO `json:"items,omitempty"`
+	Flow         Flow                       `json:"flow"`
+	ChaCompanyID string                     `json:"chaCompanyId"`
+	Items        []CreateConsignmentItemDTO `json:"items,omitempty"`
 }
 
 func (d *CreateConsignmentDTO) Validate() error {
-	if d.ChaID == "" {
-		return fmt.Errorf("chaId is required")
+	if d.ChaCompanyID == "" {
+		return fmt.Errorf("chaCompanyId is required")
 	}
 	if d.Flow != FlowImport && d.Flow != FlowExport {
 		return fmt.Errorf("flow must be IMPORT or EXPORT")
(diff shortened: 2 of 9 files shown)


backend/internal/consignment/service.go
```diff
@@ -168,7 +213,7 @@ func (s *Service) InitializeConsignmentByID(
 
 	wt := &mapping.WorkflowTemplate
 
-	if err := s.wm.StartWorkflow(ctx, consignment.ID, wt.WorkflowDefinition, globalContext); err != nil {
+	if err := s.wm.StartWorkflow(ctx, consignment.ID, wt.WorkflowDefinition, initialVars); err != nil {
 		tx.Rollback()
 		return nil, fmt.Errorf("failed to register workflow: %w", err)
 	}
```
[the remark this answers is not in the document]
sthanikan2000: Acknowledged — this is a real dual-write hazard between the Postgres commit and the Temporal `StartWorkflow` call, but the ordering predates this PR (the original two-stage flow had the same `StartWorkflow → Commit` sequence), so addressing it inside this refactor would expand scope significantly.

Tracked as a follow-up for a dedicated design pass: OpenNSW/nsw#540. The issue lays out the four candidate fixes (lean on Temporal idempotency + a resume endpoint, reverse the commit/start order + reconciler, transactional outbox, or saga with compensation) so we can pick one deliberately rather than patch it drive-by here.

backend/internal/consignment/service.go
```diff
@@ -517,20 +564,41 @@ func (s *Service) buildConsignmentDetailDTO(
...
+	raw, err := json.Marshal(record)
+	if err != nil {
+		return nil, err
+	}
+	out := make(map[string]any)
+	if err := json.Unmarshal(raw, &out); err != nil {
+		return nil, err
+	}
+	return out, nil
+}
```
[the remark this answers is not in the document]
sthanikan2000: Considered and declined for this call site — leaving the `json.Marshal`+`json.Unmarshal` round-trip and added a comment in d8fa82f2 explaining the choice.

Two reasons:

1. **Self-maintaining contract.** The round-trip ties the workflow-variable shape to the same JSON tags used by the REST response. Adding a new field on `company.Record` with a json tag flows through everywhere consistently. The hand-rolled map silently drops it until a workflow step notices, which is a fragile coupling — exactly the kind of bug that hides for a release or two.

2. **No hot path.** This runs once per Stage 2 init, not in a loop. The reflection cost is microseconds and dwarfed by the surrounding Postgres and Temporal calls. There's no measurable upside to optimize.

If `companyRecordToMap` ever does land on a hot path, a `mitchellh/mapstructure`-style decoder would preserve the self-maintaining property without the marshal cost — but that's a future change, not this PR.

Also worth noting the snippet in the suggestion has a subtle bug: leaving `record.Data` as `json.RawMessage` in the map means downstream JSON serializers will base64-encode it instead of emitting the nested object, which would break workflow steps that expect `traderCompany.data.br_no` etc.

ginaxu1 (approved):

backend/internal/consignment/model.go
```diff
@@ -72,17 +75,17 @@ type CreateConsignmentItemDTO struct {
...
-// Legacy / single-stage: provide flow + items → creates consignment and initializes workflow.
+// Stage 1 (two-stage flow): provide flow + chaCompanyId → creates shell with state INITIALIZED.
+// A specific CHA is assigned later, at Stage 2, when a CHA from chaCompanyId claims the consignment.
 type CreateConsignmentDTO struct {
-	Flow  Flow                       `json:"flow"`
-	ChaID string                     `json:"chaId"`
-	Items []CreateConsignmentItemDTO `json:"items,omitempty"`
+	Flow         Flow                       `json:"flow"`
+	ChaCompanyID string                     `json:"chaCompanyId"`
+	Items        []CreateConsignmentItemDTO `json:"items,omitempty"`
```
ginaxu1: can we remove `Items`? seems in the 2-stage refactor (Stage 1 creates the shell, Stage 2 initializes the items), the router now discards `req.Items`

backend/internal/app/bootstrap/app.go
```diff
@@ -93,8 +94,8 @@ func Build(ctx context.Context, cfg *config.Config) (*App, error) {
 		return nil, fmt.Errorf("failed to create temporal client: %w", err)
 	}
 
-	consignmentService := consignment.NewService(db, templateService, chaService, hsCodeService)
-	consignmentRouter := consignment.NewRouter(consignmentService, chaService)
+	consignmentService := consignment.NewService(db, templateService, chaService, companyService, userProfileService, hsCodeService)
```
ginaxu1, line 97: double check this circular dependency. seems `workflowruntime.NewRuntime` takes `consignmentService` as a dependency, and `consignmentService` takes `workflowRuntime.Manager()` as a dependency via `RegisterWorkflowManager`

better to decouple these domains using an Event Bus or a Go channel callback mechanism
- have workflow runtime blindly emit events (e.g. WorkflowCompletedEvent), and the consignmentService should be registered as a listener to those events
- don't need consignment service to directly hold a reference to the workflow manager

ginaxu1 (changes requested):

backend/internal/consignment/model.go
sthanikan2000: Good catch — `req.Items` was a leftover from the pre-two-stage shape. Removed both `Items` from `CreateConsignmentDTO` and the now-unused `CreateConsignmentItemDTO` in b00a4487. Build / vet / tests / lint all clean.

backend/internal/app/bootstrap/app.go
sthanikan2000: This is not related to this PR. There is no circular dependency currently.

What we implemented here is, there is two requirements:
1. WorkflowRuntime need to register upstream service we need to call to give updates
2. Upstream service needs a `workflowManger` to start workflow

So this was the simple pattern I used to solve this.
If you go through the `workflowruntime.NewRuntime` and `consignmentService.RegisterWorkflowManager` functions we're wiring the callbacks in the both sides. Not using the service reference directly.

I created a seperate issue to address this #541

ginaxu1 (approved):

Aravinda-HWK (approved):
LGTM

PR #555, opened 2026-05-25
feat(taskv2): integrate zone-based rendering end-to-end

### LINKED ISSUE #395: Move from plugin-based task execution to fully configuration-based atomic activities
**Problem**

The backend currently has two execution models living side by side:

- legacy plugin-based task types under `backend/internal/task/plugin` such as `SIMPLE_FORM`, `WAIT_FOR_EVENT`, and `PAYMENT`
- configuration-driven activities under `backend/internal/taskworkflow/activity` such as `JSON_BUILDER` and `REST_CALLER`

This split creates architectural duplication and slows down the migration toward configuration-based workflows. The plugin model still mixes business behavior, task orchestration, and rendering concerns, while the taskworkflow activity model is moving toward smaller atomic units with explicit config, inputs, and outputs.

As a result:

- workflow behavior is not fully described in configuration
- adding new flows often still requires Go code changes in plugin implementations
- rendering and execution concerns remain coupled in legacy paths
- the migration plan in `docs/migrations/configuration-based-workflows-and-rendering.md` is only partially realized
- there is no single end-state execution model for workflow-backed tasks

We need to complete the transition from monolithic plugin-based tasks to atomic configuration-driven activities so that workflows are described and composed entirely through configuration.

**Proposed Solution**

Move the system toward a fully configuration-based execution model built on atomic activities.

Key changes:

- define the taskworkflow activity layer as the primary execution model for workflow-backed tasks
- decompose current plugin-based task types (`SIMPLE_FORM`, `WAIT_FOR_EVENT`, `PAYMENT`) into smaller atomic activities with explicit config contracts
- express workflow behavior through configured activity composition rather than hardcoded task plugins
- standardize activity contracts around:
  - config-driven inputs
  - structured `Result`
  - explicit status and output mapping
  - renderer-facing payloads where needed
- move rendering concerns out of execution logic and into configuration and/or a dedicated rendering layer, consistent with `docs/migrations/configuration-based-workflows-and-rendering.md`
- introduce a migration path for existing workflows and templates so that legacy plugin-backed tasks can be replaced incrementally
- define the minimal set of atomic activities needed to replace existing plugin behavior, for example:
  - JSON data shaping
  - HTTP/API calling
  - value reduction / computed state
  - wait/callback handling
  - payment-specific atomic steps
  - form submission / validation steps
- document the config contracts for each activity and establish integration coverage for activity execution paths

Expected outcome:

- workflows become fully configuration-driven
- plugin-specific business logic is reduced or removed
- execution becomes more composable and easier to evolve
- the architecture aligns with the documented migration plan toward headless atomic activities and configuration-based rendering

**Alternatives**

- keep the current hybrid model and continue adding features to both plugins and activities
  - this is the lowest-effort path short term, but it preserves duplication and weakens the target architecture
- keep plugins as the main execution unit and only use activities for a few utility steps
  - this limits migration value and keeps rendering/execution coupling in place
- fully rewrite the task system in one step
  - this would be high risk and likely too disruptive without an incremental migration strategy

My recommendation is an incremental migration:

- treat taskworkflow activities as the target model
- migrate plugin behavior into atomic activities step by step
- keep compatibility only as a bridge, not as the long-term architecture

mushrafmim:
## Summary

Wires the new zone-based renderer end-to-end in `trader-app` behind the `VITE_USE_ZONE_RENDERER` feature flag. The backend grows a `ZoneViewAssembler` that projects a task record into a `ZoneView` payload; the frontend consumes it via a clean workspace/presentational zone split. Closes #395.

Download these files and place it inside the `backend/configs/`
[fcau-configs.zip](https://github.com/user-attachments/files/28224670/fcau-configs.zip)

## Type of Change

## Changes Made

**Backend**
- New `taskv2/renderer/zone_assembler.go` + `zoneview.go`: assembles `ZoneView` (sections, actions, state, audit) from a `TaskRecord` + its render config, using the existing uiprojector pipeline.
- `taskv2/http_handler.go`: `GET /tasks/{id}` returns the assembled `ZoneView` when the new path is active.
- `taskv2/wiring.go` + `app/bootstrap/app.go`: wires the assembler into the renderer factory.

**Frontend — zone architecture**
- `WorkspaceZone.tsx` (new): owns workspace chrome — `formRef`, validity state, submitting state, action bar. For `FORM` components, renders `FormRenderer` directly to attach the ref + validity callback; otherwise falls back to the generic dispatcher.
- `PresentationalZone.tsx` (new): dumb wrapper for `reference`/`instructions` zones — header + generic dispatcher.
- `TraderZoneLayout.tsx`: slimmed down to a thin layout that slots `WorkspaceZone` for the workspace and `PresentationalZone` for the rest. No form state. Workspace is keyed on `task_id:state` so a state transition remounts it with fresh form state.
- `renderers/index.tsx`: `renderZoneComponent(component)` reverted to a pure single-arg dispatcher.
- `renderers/FormRenderer.tsx`: exposes `FormHandle = { getData, autoFill }` via ref. `onValidityChange` callback prop emits validity on every JSONForms change. Built-in validity helper treats empty string/null/undefined/empty-array as missing for any schema-required field (works around AJV's permissive `required` check). Auto-fill encapsulates `autoFillForm` internally so the workspace doesn't reach into form internals.

**Frontend — wiring**
- `screens/TaskDetailScreen.tsx`: when `VITE_USE_ZONE_RENDERER` is on, fetches `ZoneView` via `getZoneView` and renders `TraderZoneLayout`. `onSubmitForm` POSTs the data, waits 1.5s for Temporal to advance, then refetches.
- `services/task.ts`: adds `getZoneView` + `submitTaskStep`.
- `services/api.ts`: `apiPost`/`apiPut` now treat an empty body on a 2xx response as `undefined` (was throwing). 204 No Content from `POST /tasks/{id}` is a legitimate success.

## Testing

- [x] I have tested edge cases (empty form invalidation, post-submit refetch, 204 handling, form-state reset on task state transition)
- [x] All existing tests pass

Manually exercised:
1. Submit button starts disabled on empty form, enables when all required fields filled, disables again if a required field is cleared.
2. Submit POSTs the form, receives 204, waits 1.5s, refetches — `WorkspaceZone` remounts cleanly when `task.state` transitions.
3. Auto-fill button (gated by `VITE_SHOW_AUTOFILL_BUTTON`) populates the form and enables submit.
4. Reference zone hides when there's no `reviewerform` data, appears once external review writes it back.
5. v1 rendering path (flag off) untouched — verified existing flows still work.

## Checklist

- [x] My changes generate no new warnings (`tsc --noEmit` clean; no new lint errors in changed files)

## Related Issues

Closes #395

## Additional Notes

**Architecture rationale.** The split between `WorkspaceZone` and `PresentationalZone` was deliberate. Workspace is the only *editable* zone — it owns the form-ref + validity + submit-state coupling. Reference and instructions are presentational. Keeping the generic `renderZoneComponent` dispatcher pure (single arg, no form plumbing) means new presentational renderers don't pay for workspace-only concerns.

**Known seams** (intentional, deferred until needed):
- Workspace identity is position-based (`name === 'workspace'`). Fine while we have one editable zone per task; would need a `mode` flag if a task ever needs multiple editors.
- `FORM` is special-cased inside `WorkspaceZone`. When a second submittable renderer appears (provider picker for payments — see follow-up below), the special-case generalizes to a `SubmittableHandle` interface.
- `getTaskInfo` and `getZoneView` GET the same URL with different response interpretations. Will consolidate when the v1 path is retired.

**Backend echo on refetch.** The backend returns the user's just-submitted data back in `payload.data`, so when state transitions are slow the form may look unchanged after refresh. The 1.5s delay + state-keyed remount mitigates this; a deeper fix is out of scope for this PR.

**Follow-up captured.** PAYMENT plugin + zone renderer (including provider-selection workspace and the generalization to `SubmittableHandle`) is captured as a separate workstream — not in scope here.

## Config changes (attached separately)

The FCAU render configs need a small companion change to exercise the new pipeline — not included in the pushed commits since they live under `backend/configs/fcau/`. I'll attach the full `fcau/` folder as a zip for reviewers to drop in locally and test against.

- `fcau/1-application/render.json`: added `visibleWhen: { requireDataKey: "reviewerform" }` to the `reference` section so the reviewer form only appears once the external review writes data back. Also added the workspace `submit_form` action for the `PENDING_USER` state.

## Deployment Notes

- Feature flag `VITE_USE_ZONE_RENDERER` defaults to `false`. The v1 rendering path is unchanged.
- Dev-only flag `VITE_SHOW_AUTOFILL_BUTTON` restores the demo auto-fill button in the new path.
- Reviewers: unzip the attached `fcau/` folder over `backend/configs/fcau/` to pick up the render-config changes before testing.

### DIFF
--- a/backend/internal/taskv2/http_handler.go
+++ b/backend/internal/taskv2/http_handler.go
@@ -33,6 +33,8 @@ func NewHTTPHandler(manager *orchestrator.TaskManager, store TaskFetcher, assemb
 //
 //	GET /api/v1/tasks/{id}
 func (h *HTTPHandler) HandleGetTask(w http.ResponseWriter, r *http.Request) {
+	// TODO: retrieve the authenticated context and validate it against the
+	// task's ownership bounds before returning ZoneView.
 	taskID := r.PathValue("id")
 	if taskID == "" {
 		writeJSONError(w, http.StatusBadRequest, "task id is required")
@@ -47,7 +49,8 @@ func (h *HTTPHandler) HandleGetTask(w http.ResponseWriter, r *http.Request) {
 
 	zv, err := h.Assembler.Assemble(r.Context(), record)
 	if err != nil {
-		writeJSONError(w, http.StatusInternalServerError, err.Error())
+		slog.Error("taskv2: failed to assemble zone view", "taskId", taskID, "error", err)
+		writeJSONError(w, http.StatusInternalServerError, "An internal error occurred while loading the task")
 		return
 	}
 
@@ -59,6 +62,8 @@ func (h *HTTPHandler) HandleGetTask(w http.ResponseWriter, r *http.Request) {
 //	POST /api/v1/tasks/{id}
 //	body: arbitrary JSON object — passed through to the task plugin
 func (h *HTTPHandler) HandleCompleteTaskStep(w http.ResponseWriter, r *http.Request) {
+	// TODO: retrieve the authenticated context and validate it against the
+	// task's ownership bounds before completing the step.
 	taskID := r.PathValue("id")
 
 	var payload map[string]any
@@ -87,7 +92,8 @@ func (h *HTTPHandler) HandleCompleteTaskStep(w http.ResponseWriter, r *http.Requ
 	payload = unwrapOGACallback(payload)
 
 	if err := h.Manager.CompleteTaskStep(r.Context(), taskID, payload); err != nil {
-		writeJSONError(w, http.StatusInternalServerError, err.Error())
+		slog.Error("taskv2: failed to complete task step", "taskId", taskID, "error", err)
+		writeJSONError(w, http.StatusInternalServerError, "An internal error occurred while processing the task")
 		return
 	}
 
--- a/backend/internal/taskv2/http_handler.go
+++ b/backend/internal/taskv2/http_handler.go
@@ -1,6 +1,7 @@
 package taskv2
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"io"
@@ -8,19 +9,29 @@ import (
 	"net/http"
 
 	"github.com/OpenNSW/nsw-task-flow/orchestrator"
+	tfstore "github.com/OpenNSW/nsw-task-flow/store"
+
+	"github.com/OpenNSW/nsw/internal/taskv2/renderer"
 )
 
+// TaskFetcher is the narrow surface HandleGetTask needs from the task store.
+type TaskFetcher interface {
+	GetTask(ctx context.Context, taskID string) (tfstore.TaskRecord, bool)
+}
+
 type HTTPHandler struct {
-	Manager *orchestrator.TaskManager
+	Manager   *orchestrator.TaskManager
+	Store     TaskFetcher
+	Assembler *renderer.ZoneViewAssembler
 }
 
-func NewHTTPHandler(manager *orchestrator.TaskManager) *HTTPHandler {
-	return &HTTPHandler{Manager: manager}
+func NewHTTPHandler(manager *orchestrator.TaskManager, store TaskFetcher, assembler *renderer.ZoneViewAssembler) *HTTPHandler {
+	return &HTTPHandler{Manager: manager, Store: store, Assembler: assembler}
 }
 
-// HandleGetTask returns the rendered UI payload for a single task.
+// HandleGetTask returns the ZoneView payload for a single task.
 //
-//	GET /api/v2/tasks/{id}
+//	GET /api/v1/tasks/{id}
 func (h *HTTPHandler) HandleGetTask(w http.ResponseWriter, r *http.Request) {
 	taskID := r.PathValue("id")
 	if taskID == "" {
@@ -28,18 +39,24 @@ func (h *HTTPHandler) HandleGetTask(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	info, err := h.Manager.GetTaskRenderInfo(r.Context(), taskID)
+	record, ok := h.Store.GetTask(r.Context(), taskID)
+	if !ok {
+		writeJSONError(w, http.StatusNotFound, "task not found")
+		return
+	}
+
+	zv, err := h.Assembler.Assemble(r.Context(), record)
 	if err != nil {
 		writeJSONError(w, http.StatusInternalServerError, err.Error())
 		return
 	}
 
-	writeJSONResponse(w, http.StatusOK, info)
+	writeJSONResponse(w, http.StatusOK, zv)
 }
 
 // HandleCompleteTaskStep advances a task by submitting a step payload.
 //
-//	POST /api/v2/tasks/{id}/step
+//	POST /api/v1/tasks/{id}
 //	body: arbitrary JSON object — passed through to the task plugin
 func (h *HTTPHandler) HandleCompleteTaskStep(w http.ResponseWriter, r *http.Request) {
 	taskID := r.PathValue("id")
--- a/portals/apps/trader-app/src/screens/TaskDetailScreen.tsx
+++ b/portals/apps/trader-app/src/screens/TaskDetailScreen.tsx
@@ -2,14 +2,15 @@ import { useState, useEffect, useCallback, useRef } from 'react'
 import { useParams, useNavigate } from 'react-router-dom'
 import { Button, Spinner, Text } from '@radix-ui/themes'
 import { ArrowLeftIcon } from '@radix-ui/react-icons'
-import { getTaskInfo } from '../services/task'
+import { getTaskInfo, getZoneView, submitTaskStep } from '../services/task'
 import { useApi } from '../services/ApiContext'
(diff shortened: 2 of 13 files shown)


portals/apps/trader-app/src/screens/TaskDetailScreen.tsx
```diff
@@ -131,6 +115,47 @@ export function TaskDetailScreen() {
...
+        <TraderZoneLayout
+          task={zoneView}
+          onSubmitForm={async (_command, data) => {
+            if (!taskId) return
+            await submitTaskStep(taskId, data, api)
+            // Give Temporal a moment to advance the workflow before refetching.
+            await new Promise((resolve) => setTimeout(resolve, POST_SUBMIT_REFETCH_DELAY_MS))
+            console.log('Submitted task step, refetching...')
+            await fetchTask()
+          }}
```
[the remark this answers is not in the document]
mushrafmim: Addressed

portals/apps/trader-app/src/zones/renderers/FormRenderer.tsx
```diff
@@ -1,15 +1,93 @@
...
+}
+
+// Walks the schema's `required` arrays and checks each path against the data.
+// Treats undefined, null, empty string, and empty array as "missing".
+function allRequiredFilled(schema: JsonSchema | undefined, data: unknown): boolean {
+  if (!schema || typeof schema !== 'object') return true
+  const required = (schema as { required?: string[] }).required
+  const properties = (schema as { properties?: Record<string, JsonSchema> }).properties
+
+  if (Array.isArray(required) && properties && data && typeof data === 'object') {
```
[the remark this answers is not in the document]
mushrafmim: Addressed.

backend/internal/taskv2/http_handler.go
```diff
@@ -1,45 +1,62 @@
...
+func NewHTTPHandler(manager *orchestrator.TaskManager, store TaskFetcher, assembler *renderer.ZoneViewAssembler) *HTTPHandler {
+	return &HTTPHandler{Manager: manager, Store: store, Assembler: assembler}
 }
 
-// HandleGetTask returns the rendered UI payload for a single task.
+// HandleGetTask returns the ZoneView payload for a single task.
 //
-//	GET /api/v2/tasks/{id}
+//	GET /api/v1/tasks/{id}
 func (h *HTTPHandler) HandleGetTask(w http.ResponseWriter, r *http.Request) {
```
ginaxu1, line 35: both `HandleCompleteTaskStep` & `HandleGetTask`, shouldn't we handle authorization? can you add a // TODO to retrieve the authenticated context and validate it against the Task's ownership bounds before returning ZoneView or completing steps

ginaxu1 (changes requested):

backend/internal/taskv2/http_handler.go
```diff
@@ -1,45 +1,62 @@
...
-	info, err := h.Manager.GetTaskRenderInfo(r.Context(), taskID)
+	record, ok := h.Store.GetTask(r.Context(), taskID)
+	if !ok {
+		writeJSONError(w, http.StatusNotFound, "task not found")
+		return
+	}
+
+	zv, err := h.Assembler.Assemble(r.Context(), record)
 	if err != nil {
 		writeJSONError(w, http.StatusInternalServerError, err.Error())
```
ginaxu1: better to return a generic error string to the user instead
```go
    slog.Error("failed to complete task step", "taskId", taskID, "error", err)
    writeJSONError(w, http.StatusInternalServerError, "An internal error occurred while processing the task")
```

ginaxu1 (changes requested):

backend/internal/taskv2/http_handler.go
ginaxu1, line 69: missing `defer r.Body.Close()` above line 64

ginaxu1 (changes requested):

backend/internal/taskv2/http_handler.go
mushrafmim: Skipping this one — `net/http` already closes the request body for us once the handler returns (see the `http.Request.Body` doc: *"The Server will close the request body. The ServeHTTP Handler does not need to."*). The `defer Body.Close()` convention applies to `http.Response.Body` on the **client** side, where the caller owns the lifecycle.

It would matter here if we hijacked the connection, handed `r.Body` to a goroutine that outlives the handler, or proxied the body — none of which we do. Added boilerplate without a real leak to prevent, so leaving it out to stay consistent with the rest of the codebase. Happy to revisit if you have a specific scenario in mind.

backend/internal/taskv2/http_handler.go
mushrafmim: Added.

backend/internal/taskv2/http_handler.go
mushrafmim: Added.

ginaxu1 (approved):

sthanikan2000 (approved):
lgtm

PR #557, opened 2026-05-26
feat: add manager and sms.

maneesha-xyz:
## Summary

Replaces the legacy `backend/pkg/notification` package with a new `backend/pkg/notifications` package that introduces a `Manager` + pluggable `Provider` model. Adds Email and SMS providers backed by HTTP APIs, driven by a JSON config file (`NOTIFICATIONS_CONFIG_PATH`). Bootstrap is rewired to construct the new manager from config rather than hard-wiring channels.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

Breaking: `App.NotificationManager` renamed to `App.NotificationsManager`; old `pkg/notification` package removed. Callers must migrate to the new manager/provider API.

## Changes Made

- **New `backend/pkg/notifications/` package**
  - `manager.go` — `Manager` routes `Request`s to registered `Provider`s keyed by `ChannelType`. Loads per-provider config from a JSON file at construction; fails fast if a provider's key is missing or its `Configure` returns an error.
  - `providers/email.go` — HTTP-based `EmailProvider` (bearer token auth via `pkg/remote/auth`).
  - `providers/sms.go` — HTTP-based `SMSProvider`.
  - `types.go` — `Request`, `ChannelType`, `Provider` interface, `Config` + `ErrConfigPathRequired`.
  - Unit tests for the manager (happy path, missing config key, configure failure, invalid path, bad JSON, duplicate channel) and integration test for the SMS provider.
- **Removed legacy `backend/pkg/notification/`** — manager, channels (`email`, `gov_sms`), tests, integration test data, README. All replaced by the new package.
- **Bootstrap (`internal/app/bootstrap/app.go`)**
  - Constructs `notifications.Manager` from `cfg.Notification.ConfigPath` with `EmailProvider` and `SMSProvider`. Logs a warning (does not fail startup) when notifications are unavailable.
  - Renames `App.NotificationManager` → `App.NotificationsManager`.
- **Config (`internal/config/config.go`)** — adds `NotificationConfig.ConfigPath`, sourced from `NOTIFICATIONS_CONFIG_PATH` (defaults to `configs/notifications.json`).
- Unrelated changes pulled in via merges from `main`: taskv2 registry / renderer wiring, template service registry, consignment service tweaks, `start-dev.sh` simplification, `nsw-srilanka` migration.

## Testing

- [x] All existing tests pass

Unit tests in `backend/pkg/notifications/manager_test.go` cover validate failure, missing provider config, configure error, bad JSON, duplicate-channel last-wins, unknown channel, and provider send failure. SMS provider integration test in `backend/pkg/notifications/test/integration/sms_integration_test.go`.

Run: `cd backend && make test`.

## Checklist

## Screenshots/Demo

N/A — backend-only refactor.

## Deployment Notes

- Set `NOTIFICATIONS_CONFIG_PATH` (defaults to `configs/notifications.json`). The file must contain JSON objects keyed by channel (`email`, `sms`) with provider-specific fields (`baseURL`, `token`).
- Example:
  ```json
  {
    "email": { "baseURL": "https://email.api.example", "token": "..." },
    "sms":   { "baseURL": "https://sms.api.example",   "token": "..." }
  }
  ```
- If the file is missing or malformed at startup, notifications are disabled (warning logged) but the server still boots.


backend/pkg/notifications/providers/sms.go
```diff
@@ -0,0 +1,76 @@
...
+	if s.cfg.UserName == "" {
+		return errors.New("sms provider: username is required")
+	}
+	if s.cfg.Password == "" {
+		return errors.New("sms provider: password is required")
+	}
+	return nil
+}
+
+func (s *SMSProvider) Send(ctx context.Context, req notifications.Request) error {
```
ginaxu1, line 67: in [types.go](https://github.com/maneeshaxyz/nsw/blob/feat/notification-manager-and-providers/backend/pkg/notifications/types.go#L29), `Validate` method is defined with a pointer receiver... so shouldn't this be `Send(ctx context.Context, req *Request) error`? 

Also is `backend/pkg/notification` (singular)  Legacy? can you delete?

ginaxu1 (changes requested):

backend/pkg/notifications/providers/sms.go
```diff
@@ -0,0 +1,82 @@
...
+	if s.client == nil {
+		return errors.New("sms provider not configured")
+	}
+	if err := s.client.JSONRequest(ctx, remote.Request{
+		Method: http.MethodPost,
+		Path:   "/govsms/V1/prod/send",
```
ginaxu1: are you confident it's okay to hardcode this? what if version changes (V2) - should we instead include the path entirely in the BaseURL configuration so the client only needs to specify /send or an empty path?

ginaxu1 (changes requested):

backend/pkg/notifications/manager.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("load notification config: %w", err)
+	}
+
+	m := &Manager{
+		providers: make(map[ChannelType]Provider, len(providers)),
+	}
+
+	for _, p := range providers {
+		raw, ok := cfgMap[string(p.Type())]
```
ginaxu1, line 37: what happens if multiple providers of the same ChannelType are passed into NewManager? seems current logic would overwrite previous provider. 

better to catch that and return error
```go
if _, dup := m.providers[p.Type()]; dup {
    return nil, fmt.Errorf("duplicate notification provider registered for channel: %q", p.Type())
}
```

ginaxu1 (changes requested):

backend/pkg/notifications/manager.go
```diff
@@ -0,0 +1,64 @@
+package notifications
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+)
```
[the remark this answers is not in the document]
maneesha-xyz: It compiles and is imported since its there in the same folder.

backend/pkg/notifications/types.go
ginaxu1, line 21: consideration for future: refactor this to have generic Payload, otherwise we risk bloating this `Request` struct with channel-specific variables down the line like

suggestion: 
```
type Request struct {
Channel ChannelType `json:"channel"`
    To      string      `json:"to"`
    Payload map[string]any `json:"payload"` // allows specific providers to extract what they need
}
```

ginaxu1 (changes requested):

backend/pkg/notifications/providers/sms.go
maneesha-xyz: Yeah, this should change. I wanted to test the e2e flow and this got left out. Will change, thanks.

backend/pkg/notifications/types.go
maneesha-xyz: Yeah, I think this makes sense.

backend/pkg/notifications/providers/sms.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (s *SMSProvider) Type() notifications.ChannelType { return notifications.ChannelSMS }
+
+func (s *SMSProvider) Configure(raw json.RawMessage) error {
+	if err := json.Unmarshal(raw, &s.cfg); err != nil {
+		return fmt.Errorf("unmarshal sms config: %w", err)
+	}
+	if s.cfg.BaseURL == "" {
+		return errors.New("baseURL is required")
+	}
```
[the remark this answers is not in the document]
maneesha-xyz: fixed.

backend/pkg/notifications/providers/email.go
```diff
@@ -0,0 +1,71 @@
...
+func (e *EmailProvider) Type() notifications.ChannelType { return notifications.ChannelEmail }
+
+func (e *EmailProvider) Configure(raw json.RawMessage) error {
+	var cfg emailConfig
+	if err := json.Unmarshal(raw, &cfg); err != nil {
+		return fmt.Errorf("unmarshal email config: %w", err)
+	}
+	if cfg.BaseURL == "" {
+		return errors.New("baseURL is required")
+	}
```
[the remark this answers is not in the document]
maneesha-xyz: fixed

backend/pkg/notifications/manager.go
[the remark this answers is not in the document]
maneesha-xyz: resolved.

backend/pkg/notifications/providers/sms.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+
+	"github.com/OpenNSW/nsw/backend/pkg/notifications"
+	"github.com/OpenNSW/nsw/backend/pkg/remote"
+)
```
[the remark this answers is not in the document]
maneesha-xyz: Uneeded.

ginaxu1 (changes requested):
make sure all outbound HTTP requests utilize the passed context, e.g. 
```
req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, body)
``` 

This should be the same context.Context passed into `Send()` so that network calls respects upstream context cancellations or timeouts

backend/pkg/notifications/providers/email.go
```diff
@@ -0,0 +1,74 @@
...
+	if cfg.Token == "" {
+		return errors.New("token is required")
+	}
+	e.client = remote.NewClient(cfg.BaseURL, remote.WithAuthenticator(remoteauth.NewBearer(remoteauth.BearerConfig{Token: cfg.Token})))
+	return nil
+}
+
+func (e *EmailProvider) Send(ctx context.Context, req notifications.Request) error {
+	if e.client == nil {
```
ginaxu1, line 58: for `Provider` interface, why use a 2-phase initialization pattern (creating an empty struct and then calling a separate `Configure()` method) instead of parsing the configuration first and passing it directly into the constructor?

if you pass the config upfront, then guarantees provider is fully initialized and valid upon creation (and you could safely remove these `if e.client == nil` checks at the start of every `Send` method)

ginaxu1 (changes requested):

ginaxu1 (approved):

sthanikan2000 (approved):

PR #561, opened 2026-05-26
feat(authz): introduce Principal abstraction and RequireScope middleware (v1 plumbing)

sthanikan2000:
## Summary

- Adds a new `internal/authz` package that complements `internal/auth`: a `Principal` interface unifying user/client principal types and a `Manager.RequireScope(scope)` middleware factory for coarse authorization gating per route.
- Scopes are resolved in-package (static `role -> scopes` map for users, static `client_id -> scopes` allowlist for M2M) so this works today without any IdP / Thunder change. The interface is a seam — when Thunder issues real scope claims later, swap the resolver and callers don't change.
- Pure plumbing — **no existing handler, route, or service is touched** and **no behavioural change**. v1 is unreferenced (no caller yet) until the first feature migrates to it in a follow-up PR.

## Why

Current authz patterns in the repo are inconsistent and partially broken:
- Every handler repeats `if authCtx == nil || authCtx.User == nil { 401 }`, which silently locks out M2M callers from endpoints that should serve them.
- [`internal/consignment/router.go:85`](https://github.com/OpenNSW/nsw/blob/main/backend/internal/consignment/router.go#L85) carries an explicit `TODO: Proper AuthZ` and accepts a `role=trader|cha` **query parameter** from the caller — a critical flaw, since the role isn't enforced from JWT claims.
- Service-layer ownership checks (e.g. [`consignment/service.go:160`](https://github.com/OpenNSW/nsw/blob/main/backend/internal/consignment/service.go#L160)) are ad-hoc and inconsistent across services.

This PR establishes the foundational primitives so authorization can be done the same way everywhere, with the policy decisions living **inside the downstream service** — required because the same endpoint can serve frontend users and M2M clients with different rules.

## Architecture (layered hybrid)

| Layer | Responsibility | Lives in |
|---|---|---|
| Authentication middleware (exists) | Parse JWT, build `AuthContext`. | `internal/auth/middleware.go` |
| **`RequireScope` middleware (new)** | Reject tokens missing a declared scope. Coarse gate, declared per route. | `internal/authz/middleware.go` |
| Handler (per-feature, future) | Pull `Principal` from context, forward to service. Maps `ErrForbidden` -> 403. | `**/router.go` |
| Service (per-feature, future) | All fine-grained authz: ownership, role branching, etc. Receives `Principal`. | `**/service.go` |

## What's in v1 (this PR)

| File | Purpose |
|---|---|
| `principal.go` | `Principal` interface, `Kind` enum, `ErrUnauthenticated` / `ErrForbidden` sentinels |
| `principal_impl.go` | Implementation wrapping `*auth.AuthContext`; scope resolution delegated to the `Manager` |
| `config.go` | `Config{RoleScopes, ClientScopes}` + `Validate` (empty config is valid) |
| `manager.go` | `Manager`, `NewManager`, `Principal(ctx) (Principal, bool)` |
| `middleware.go` | `Manager.RequireScope(scope)` — 401 anonymous / 403 missing-scope, JSON shape matches `auth` |
| `principal_test.go`, `middleware_test.go` | 13 tests / 18 subtests covering user, client, anonymous, unknown role, unmapped client, map-isolation, JSON response shape |

## Deferred (future PRs)

- **v2 — first migration:** Migrate `internal/consignment` onto the new pattern. Replace the `role=` query parameter with `p.HasRole(...)` branching inside the service. Wire `RequireScope("consignments:read"|"consignments:write")` on the routes in `bootstrap/app.go`. **This is the PR that closes the existing `TODO: Proper AuthZ`.**
- **v3 — broader rollout:** Migrate remaining handlers (tasks, hscodes, profiles, storage) one feature at a time. Stop hardcoding `authCtx.User == nil` so M2M can reach these endpoints where appropriate.
- **v4 — security hardening:** Secure the unauthenticated payment webhook (`payments/http_handler.go:44`), audit-log every `ErrForbidden`, add 401/403 metrics.
- **Deferred indefinitely:** JWT `scope` claim from Thunder (swap the resolver impl when ready; callers don't change), group/role hierarchy resolver, heavy policy engines (OPA/Casbin).

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `golangci-lint run ./internal/authz/...` — 0 issues
- [x] `go test ./internal/authz/...` — 13 tests / 18 subtests passing
- [x] No existing file modified (diff vs `HEAD` shows only files added under `internal/authz/`)
- [ ] Reviewer: confirm `Principal` interface shape is what v2 will need
- [ ] Reviewer: confirm "scopes derived from static config maps" is acceptable as the initial source (vs. waiting for Thunder to issue JWT scope claims)
- [ ] CI green

### DIFF
--- a/backend/internal/authz/manager.go
+++ b/backend/internal/authz/manager.go
@@ -3,6 +3,7 @@ package authz
 import (
 	"context"
 	"fmt"
+	"slices"
 
 	"github.com/OpenNSW/nsw/internal/auth"
 )
@@ -56,9 +57,7 @@ func cloneScopeMap(src map[string][]string) map[string][]string {
 	}
 	dst := make(map[string][]string, len(src))
 	for k, v := range src {
-		copied := make([]string, len(v))
-		copy(copied, v)
-		dst[k] = copied
+		dst[k] = slices.Clone(v)
 	}
 	return dst
 }
--- a/backend/internal/authz/manager.go
+++ b/backend/internal/authz/manager.go
@@ -1,0 +1,58 @@
+package authz
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/OpenNSW/nsw/internal/auth"
+)
+
+// Manager owns the static authorization configuration and exposes helpers
+// for building Principals and constructing scope-checking middleware. The
+// manager mirrors the pattern of auth.Manager: construct once at startup
+// and inject the same instance into handlers and services that need it.
+type Manager struct {
+	roleScopes   map[string][]string
+	clientScopes map[string][]string
+}
+
+// NewManager constructs a Manager from the given configuration. Returns an
+// error if the configuration is structurally malformed. An empty config is
+// permitted; the resulting manager treats every scope check as false.
+func NewManager(cfg Config) (*Manager, error) {
+	if err := cfg.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid authz config: %w", err)
+	}
+	return &Manager{
+		roleScopes:   cloneScopeMap(cfg.RoleScopes),
+		clientScopes: cloneScopeMap(cfg.ClientScopes),
+	}, nil
+}
+
+// Principal returns the authenticated principal for the given request
+// context, resolved from the auth context injected by auth.Middleware.
+// ok=false when no auth context is attached (anonymous request) or when
+// the attached context is empty (neither user nor client present).
+func (m *Manager) Principal(ctx context.Context) (Principal, bool) {
+	authCtx := auth.GetAuthContext(ctx)
+	if authCtx == nil {
+		return nil, false
+	}
+	if authCtx.User == nil && authCtx.Client == nil {
+		return nil, false
+	}
+	return &principal{authCtx: authCtx, manager: m}, true
+}
+
+func cloneScopeMap(src map[string][]string) map[string][]string {
+	if len(src) == 0 {
+		return map[string][]string{}
+	}
+	dst := make(map[string][]string, len(src))
+	for k, v := range src {
+		copied := make([]string, len(v))
+		copy(copied, v)
+		dst[k] = copied
+	}
+	return dst
+}
--- a/backend/internal/authz/manager.go
+++ b/backend/internal/authz/manager.go
@@ -31,9 +31,15 @@ func NewManager(cfg Config) (*Manager, error) {
 
 // Principal returns the authenticated principal for the given request
 // context, resolved from the auth context injected by auth.Middleware.
-// ok=false when no auth context is attached (anonymous request) or when
-// the attached context is empty (neither user nor client present).
+// ok=false when no auth context is attached (anonymous request), when
+// the attached context is empty (neither user nor client present), or
+// when the receiver itself is nil (defensive — a nil Manager indicates
+// a wiring bug, but treating it as "no principal" avoids a delayed
+// panic on the first HasScope call).
 func (m *Manager) Principal(ctx context.Context) (Principal, bool) {
+	if m == nil {
+		return nil, false
+	}
 	authCtx := auth.GetAuthContext(ctx)
 	if authCtx == nil {
 		return nil, false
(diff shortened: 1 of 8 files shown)


backend/internal/authz/middleware.go
```diff
@@ -0,0 +1,53 @@
...
+// closure that panics on the first request.
+func (m *Manager) RequireScope(scope string) func(http.Handler) http.Handler {
+	if m == nil {
+		panic("authz: RequireScope called on a nil Manager")
+	}
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			p, ok := m.Principal(r.Context())
+			if !ok {
+				slog.Debug("authz: no principal on request", "scope", scope, "path", r.URL.Path)
```
ginaxu1: pass context (request trace) 
```go
slog.DebugContext(r.Context(), "authz: no principal on request", ...)
```

backend/internal/authz/middleware.go
```diff
@@ -0,0 +1,53 @@
...
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusUnauthorized)
+				_, _ = w.Write([]byte(`{"error":"unauthorized","message":"authentication required"}`))
+				return
+			}
+			if !p.HasScope(scope) {
+				slog.Warn("authz: principal lacks required scope",
```
ginaxu1: same here, pass context (request trace)
```go
slog.WarnContext(r.Context(), ... 
```

backend/internal/authz/manager.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return &principal{authCtx: authCtx, manager: m}, true
+}
+
+func cloneScopeMap(src map[string][]string) map[string][]string {
+	if len(src) == 0 {
+		return map[string][]string{}
+	}
+	dst := make(map[string][]string, len(src))
+	for k, v := range src {
```
ginaxu1, line 59: rewrite using `slices.Clone`
```go
for k, v := range src {
    dst[k] = slices.Clone(v)
}
```

backend/internal/authz/middleware.go
```diff
@@ -0,0 +1,53 @@
...
+					"scope", scope,
+					"kind", p.Kind(),
+					"subject", p.Subject(),
+					"path", r.URL.Path,
+				)
+				w.Header().Set("Content-Type", "application/json")
+				w.WriteHeader(http.StatusForbidden)
```
ginaxu1, line 46: don't we need security headers here? add above line 46
```go
w.Header().Set("Content-Type", "application/json")
w.Header().Set("X-Content-Type-Options", "nosniff")
```

ginaxu1 (changes requested):

backend/internal/authz/middleware.go
sthanikan2000: Addressed in 49bedca7 — switched to `slog.DebugContext(r.Context(), ...)`.

backend/internal/authz/middleware.go
sthanikan2000: Addressed in 49bedca7 — same change applied here with `slog.WarnContext(r.Context(), ...)`.

backend/internal/authz/manager.go
sthanikan2000: Addressed in 49bedca7 — rewritten as `dst[k] = slices.Clone(v)`.

backend/internal/authz/middleware.go
sthanikan2000: The 401/403 bodies emitted here are hardcoded JSON literals (`{"error":"unauthorized","message":"authentication required"}` and `{"error":"forbidden","message":"insufficient scope"}`) with no user-controlled content, so MIME-sniffing isn't an actual XSS vector today. Also, the adjacent `auth.Middleware` writes JSON 401/500 responses without this header, so adding it only here would create inconsistency between two adjacent middlewares.

Tracking this for the v4 security-hardening sweep (where we secure the payment webhook and audit-log forbidden decisions) — at that point the nosniff header goes on both middlewares together. Happy to revisit if you'd prefer it included in this PR.

ginaxu1 (approved):
Lgtm

PR #562, opened 2026-05-27
refactor(api): standardise paginated list envelope and implement real pagination on /companies

saknarajapakshe:
## Summary

Introduces a shared generic pagination type (`utils.Page[T]`) to replace duplicate per-package `ListResult` structs, wires real `offset`/`limit` pagination into the company and HS code endpoints, and optimises all list queries to skip the separate `COUNT(*)` when the result set is smaller than the page size.

## Type of Change

## Changes Made

- **`utils.Page[T]` generic envelope** — added to `backend/utils/pagination.go` along with a `NewPageResult[T]` constructor that guarantees `Items` is never `nil`; all existing `ListResult` structs in `hscode`, `consignment`, and `company` are now type aliases for this shared type
- **`utils.ParsePaginationParams`** — extracted HTTP query-param parsing (`offset`/`limit`) into a shared util so handlers no longer duplicate the parse/validate logic
- **Company endpoint pagination** — `company.ListFilter` gains `Offset`/`Limit` fields; `Service.ListCompanies` signature changes from `[]Record` to `*ListResult`; the handler now calls `ParsePaginationParams` and passes the values through; the service projects directly to `Summary` via a `SELECT id, name, has_cha` scan instead of fetching full `Record` rows and converting in the handler
- **Count-after-find optimisation** — all three services (company, hscode, consignment) now run the data query first and only issue a `COUNT(*)` when the returned page is full *and* offset > 0; short or first-page results derive the total from `len(items)`, saving a round-trip in the common case
- **`TotalCount` → `Total` field rename** — normalises the JSON field name across all list responses to match the existing company envelope contract
- **GORM query isolation fix** — all services now build a fresh `*gorm.DB` chain via a closure (`baseQuery()`) so `LIMIT`/`OFFSET` applied to the list query cannot leak into a subsequent `COUNT` query via GORM's shared clauses map
- **Frontend types unified** — `PaginatedResponse<T>` is the single shared interface in `common.ts`; `ConsignmentListResult`, `CompanyListResult`, and `HSCodeListResult` are now type aliases; removed the stale unused `PaginatedResponse` from `api.ts`; updated `ConsignmentScreen`, `PreconsignmentScreen`, and `preConsignment.ts` to use the renamed `total` field

## Testing

- [x] All existing tests pass

## Checklist

## Related Issues

Closes #534

## Additional Notes

The `TotalCount` → `Total` rename is a **breaking change to the JSON contract** for the consignment and HS code list responses. Frontend consumers within this repo are updated in the same commit; any external consumers will need to update their field reference.

The count-after-find optimisation changes the SQL query order (list query executes before count query). All affected tests have been updated to match the new execution order.


backend/internal/consignment/service.go
```diff
@@ -316,12 +316,8 @@ func (s *Service) listConsignmentsWithBaseQuery(ctx context.Context, baseQuery *
...
 	if totalCount == 0 {
-		return &ListResult{
-			TotalCount: 0,
-			Items:      []SummaryDTO{},
-			Offset:     finalOffset,
-			Limit:      finalLimit,
-		}, nil
+		result := utils.NewPageResult([]SummaryDTO{}, int64(0), finalOffset, finalLimit)
+		return &result, nil
 	}
```
sthanikan2000: Having this in two places inside a function seems redundant. Can you verify the logic implemented here is optimistic?

backend/internal/profile/company/handler.go
```diff
@@ -32,25 +34,18 @@ func (h *Handler) HandleGetCompanies(w http.ResponseWriter, r *http.Request) {
 		filter.Name = &name
 	}
 
-	records, err := h.svc.ListCompanies(r.Context(), filter)
+	offset, limit, err := utils.ParsePaginationParams(r)
 	if err != nil {
-		http.Error(w, "failed to retrieve companies", http.StatusInternalServerError)
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
sthanikan2000: Should not expose `err.Error()` in the http response

backend/internal/hscode/router.go
```diff
@@ -21,37 +22,24 @@ func NewRouter(service *Service) *Router {
...
-	if offsetStr := r.URL.Query().Get("offset"); offsetStr != "" {
-		offset, err := strconv.Atoi(offsetStr)
-		if err != nil {
-			http.Error(w, "invalid 'offset' query parameter, must be an integer", http.StatusBadRequest)
-			return
-		}
-		filter.Offset = &offset
+	offset, limit, err := utils.ParsePaginationParams(r)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
sthanikan2000: `err.Error()` should not be exposed in the http response

backend/internal/profile/company/service.go
```diff
@@ -81,12 +83,27 @@ func (s *service) ListCompanies(ctx context.Context, filter ListFilter) ([]Recor
...
+	if err := q.Order("name ASC").Offset(finalOffset).Limit(finalLimit).Find(&records).Error; err != nil {
 		slog.Error("failed to list company records", "error", err)
 		return nil, fmt.Errorf("database query failed: %w", err)
 	}
-	return records, nil
+
+	items := make([]Summary, 0, len(records))
+	for _, r := range records {
+		items = append(items, Summary{ID: r.ID, Name: r.Name, HasCHA: r.HasCHA})
+	}
```
sthanikan2000: Should optimize this. We can write the database query such that, only get required attributes. We should not write a loop for this.

portals/apps/trader-app/src/services/types/common.ts
sthanikan2000, line 1: Can you check and verify all the ListResults using this interface?

sthanikan2000 (changes requested):

backend/internal/consignment/service.go
saknarajapakshe: The totalCount == 0 short-circuit (lines 318-321) skips several downstream DB queries — node counts, end nodes, HS code batch load, and DTO construction. 
Even though we removed the 0-count case, it still works correctly but will  run unnecessary queries on empty result sets.
_So my opinion is to keep it as it is._

sthanikan2000 (dismissed):
LGTM!

backend/internal/consignment/model.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	"github.com/OpenNSW/nsw/backend/internal/hscode"
 	"github.com/OpenNSW/nsw/backend/internal/workflow/model"
+	"github.com/OpenNSW/nsw/backend/utils"
```
mushrafmim: We currently have a top-level `utils` folder that exists solely for the `pagination.go` implementation. I propose moving this into a more appropriate package structure, such as `pkg/http/pagination.go`. Since this is part of an older implementation, I think this is a good opportunity to refactor and improve the project structure.

backend/pkg/pagination/pagination.go
mushrafmim, line 1: We currently have a top-level `utils` folder that exists solely for the `pagination.go` implementation. I propose moving this into a more appropriate package structure, such as `pkg/http/pagination.go`. Since this is part of an older implementation, I think this is a good opportunity to refactor and improve the project structure.

backend/utils/pagination.go
mushrafmim, line 51: It seems like, these two functions doing the overlapping stuff, I feel like, we should make the `ParsePaginationParams` private and use it inside the `GetPaginationParams` function.

And I can see that service layer in `profile/company` is again decoding the paginated params, which is not at all required.

I feel like the service layer should not be agnostic of any pagination stuff, if it is in the filter, it should be applied.

mushrafmim (changes requested):

backend/pkg/pagination/pagination.go
saknarajapakshe: > We currently have a top-level `utils` folder that exists solely for the `pagination.go` implementation. I propose moving this into a more appropriate package structure, such as `pkg/http/pagination.go`. Since this is part of an older implementation, I think this is a good opportunity to refactor and improve the project structure.

I have a question.
If we name our package as `http` , the files those want to use` ParsePaginationParams` also need Go's standard `net/http` package.  Because of that, there will be a naming collision with `http` package. So, we will need to use an alias/renaming for one of them in every single file that uses it. 

Instead of that, I suggest naming it something like `pkg/pagination/`.

What do you think about this? @mushrafmim @sthanikan2000

sthanikan2000: for me keep the package name anything that make sense than the `http` fine

mushrafmim: Go ahead with `pkg/pagination/`

backend/utils/pagination.go
```diff
@@ -9,6 +9,22 @@ import (
...
+func NewPageResult[T any](items []T, total int64, offset, limit int) Page[T] {
+	if items == nil {
+		items = []T{}
+	}
+	return Page[T]{Items: items, Total: total, Offset: offset, Limit: limit}
+}
+
 // GetPaginationParams calculates the offset and limit for pagination based on the provided values.
 // If offset or limit are nil, default values are used. The limit is capped at a maximum value.
 func GetPaginationParams(offset *int, limit *int) (int, int) {
```
mushrafmim: This is not `GetPaginationParams` which is misleading, instead it should be `ResolvePaginationParams`,

mushrafmim (changes requested):

backend/utils/pagination.go
mushrafmim: I am backing the current implementation after thorough analysis, both API Layer and Service Layer should be doing validation (service layer should be self contained), where the API Layer mainly extracts and doing a basic level validation, while the service layer doing the proper validation.

We should remove the following validation in the ParsePaginationParams function. API Layer shouldn't know the maximum limit but should verify whether it is valid `limit > 0`.
```go
if limitVal > pageSizeMax {
    return nil, nil, fmt.Errorf("invalid 'limit' query parameter, maximum allowed is %d", pageSizeMax)
}
```
And we should rename the `GetPaginationParams` to `ResolvePaginationParams` because, already the pagination params are passed to the `GetPaginationParams` which is misleading.

sthanikan2000 (approved):
lgtm

mushrafmim (approved):
LGTM!

PR #573, opened 2026-05-29
feat(taskv2): unify zone rendering — derived interactivity, single Zone component

### LINKED ISSUE #395: Move from plugin-based task execution to fully configuration-based atomic activities

mushrafmim:
## Summary

Refactors the task rendering architecture to remove magic-string-based behavioral dispatch in the frontend and consolidate interactivity into a single, derived rule. State owns which commands are legal, sections own the handles they claim, and renderers own their own element catalog and footer chrome. The frontend collapses into a single unified `Zone` component — no `WorkspaceZone` / `PresentationalZone` split — and the wire shape ships a flat `{type, handles, payload}` per zone.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement

## Changes Made

**Backend (`internal/taskv2/renderer`)**

- `zoneview.go`: collapsed `Action` / `HandleClaim` to a single `command` field each (dropped `kind`, `variant`, alternate `action` field). `EnrichedComponent` now carries `{type, handles, payload}` — no top-level `actions`, no `role`. Section `id` field removed (map key is identity).
- `zone_assembler.go`: `mergeView` now decorates each slot with the filtered handle list (claims whose command is legal in the current state survive); role derivation lifted to the frontend.
- `renderer.go`: adapted to `nsw-task-flow`'s new `Renderer` contract returning `json.RawMessage`; internal `uiComponent` type retained; MARKDOWN payload wrapping preserved.

**Backend (`pkg/uiprojector`)**

- `blueprint.go` / `assembler.go`: dropped `ID` field from `SectionBlueprint` and `Section` (map key is identity); error messages reference loop key instead.

**Frontend (`portals/apps/trader-app/src/zones`)**

- Deleted `WorkspaceZone.tsx` and `PresentationalZone.tsx`; introduced unified `Zone.tsx`.
- `types.ts`: `Handle = {command, label, element?}`; `ZoneComponent` is a discriminated union over `FORM | MARKDOWN | REDIRECT`, each carrying optional `handles[]`. No top-level `actions` on `ZoneView`.
- `renderers/FormRenderer.tsx`: owns `FORM_ELEMENT_CATALOG` (`primary_action`, `secondary_action`, `danger_action`), its own sticky `FormActionBar`, and `HandleButton`. Interactivity is derived: editable iff `(handles?.length ?? 0) > 0 && onAction !== undefined`. The same rule sets `JsonForms`'s `readonly` flag.
- `renderers/index.tsx`: `renderZoneComponent` accepts a uniform `RenderOptions` carrying `onAction`; no type-specific branching at the call site.
- `TraderZoneLayout.tsx`: dropped role-based component selection; retains `ZONE_ORDER` for display ordering only.

**Configs (`backend/configs/fcau/*/render.json`)**

- Migrated all eight task configs: no `id` on sections, no `role`, sections may declare `handles: [{command, label, element}]`, states declare `actions: [{command}]` only.

## Testing

- [x] All existing tests pass

`go test ./internal/taskv2/... ./pkg/uiprojector/...` passes. `tsc --noEmit` clean in trader-app. The pre-existing `TestMarkdownProjector_Project` failure in `pkg/uiprojector/projector_test.go` is unrelated to this change.

## Checklist

## Related Issues

Closes #395

## Screenshots/Demo

N/A — behavior is unchanged for the existing FCAU flows; refactor is structural.

## Additional Notes

- **Wire-level breaking change.** Clients that read a top-level `actions[]` from `GET /api/v1/tasks/{id}` or a per-section `role` field will need to update — operations now ship inside each zone's `handles[]`, and a zone is "interactive" iff it has at least one legal handle.
- **Element identifiers are renderer-owned.** Adding a new visual treatment (e.g. a `tertiary_action` button style) is a renderer-side change; the wire passes only the `element` string.
- A natural follow-up is to lift `ZoneRendererProps<T>` to a `{component, onAction}` shape so every renderer (not just FORM) has the same dispatch contract — currently `FormRenderer` declares `handles?`/`onAction?` inline. Left out of this PR to keep scope tight.

## Deployment Notes

No migrations. Frontend and backend must deploy together because the wire shape changed (removed `actions`, `role`; sections gained `handles`).


sthanikan2000 (approved):
lgtm
Verified Running locally as well

sthanikan2000:
lgtm

PR #574, opened 2026-05-29
feat(blobsource): make pkg/ importable; add manifest-path + local-manifest modes

sthanikan2000:
## Summary

`pkg/blobsource` and `pkg/storage` both transitively import `internal/validation`. Go's internal-package rule blocks any module outside the `github.com/OpenNSW/nsw/backend` import tree from importing them, so these public `pkg/` packages can't be `go get`-ed from downstream projects (e.g. the agency portal, which currently has to vendor a copy of blobsource).

This PR relocates `validation` out of `internal/` so both packages become externally importable, and adds two general-purpose blobsource capabilities that downstream consumers need.

## Type of Change

## Changes Made

**Make `pkg/` packages importable**
- `git mv backend/internal/validation → backend/pkg/validation` and update the 5 import sites (`internal/auth`, `internal/config`, `internal/temporal`, `pkg/blobsource`, `pkg/storage`). Package name and API unchanged; import path only.

**Extend blobsource (all additive, backward compatible)**
- `GitHubConfig.ManifestPath` (env `BLOBSOURCE_GITHUB_MANIFEST_PATH`, default `manifest.json`) — serve a manifest from a subdirectory, e.g. `fcau/manifest.json`. `byId` entries resolve **relative to the manifest's directory**, so a repo can carry one manifest per logical group. Rejects `..` / absolute paths.
- **Local manifest mode** — `NewLocal` auto-detects a `manifest.json` in the directory and serves blobs from its `byId` map (mirroring the GitHub backend), so a local clone of a manifest-based repo with nested blobs works without flattening. Rejects path traversal. Flat-directory behavior is unchanged when no manifest is present.
- README: document the above; also fix the stale `import` path left over from the `backend/` module move (`nsw/pkg/...` → `nsw/backend/pkg/...`).

## Testing

- [x] I have added unit tests for new functionality (ManifestPath + traversal rejection, local-manifest mode + traversal/missing-file/bad-JSON cases)
- [x] All existing tests pass

`go build ./...` and `go vet ./...` clean. `go test ./...` passes for all packages touched (blobsource, validation, storage, config, auth, temporal). One unrelated pre-existing flaky test in `internal/consignment` surfaces only under full-suite parallel runs (passes isolated and repeated on both `main` and this branch; the package imports neither changed package).

## Additional Notes

No behavior change for existing configs: `Type` still accepts `local`/`github` exactly as before, and the new `Config`/`GitHubConfig` fields are optional with backward-compatible defaults. There are no current callers of `NewFromConfig` in this repo, so the surface-area risk is limited to the validation import-path move, which is mechanical.

Motivation: lets the agency portal drop its vendored copy of blobsource and depend on this package directly via `go get`. "Disabled / no source" is intentionally left as a consumer concern (represented however the app likes, e.g. a nil source) rather than baked into this package.


backend/pkg/blobsource/github.go
```diff
@@ -86,22 +96,35 @@ func NewGitHub(ctx context.Context, cfg GitHubConfig) (Source, error) {
 	if _, err := url.Parse(baseURL); err != nil {
 		return nil, fmt.Errorf("blobsource: invalid BaseURL %q: %w", baseURL, err)
 	}
+	manifestPath := cfg.ManifestPath
+	if manifestPath == "" {
+		manifestPath = "manifest.json"
+	}
+	if path.IsAbs(manifestPath) || strings.HasPrefix(manifestPath, "../") || strings.Contains(manifestPath, "/../") {
+		return nil, fmt.Errorf("blobsource: ManifestPath %q must be repo-relative and must not contain ..", manifestPath)
+	}
```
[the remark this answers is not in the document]
sthanikan2000: Addressed

backend/pkg/blobsource/local_manifest.go
```diff
@@ -0,0 +1,71 @@
...
+	if filepath.IsAbs(rel) {
+		return "", fmt.Errorf("manifest path %q must be relative", rel)
+	}
+	cleaned := filepath.Clean(filepath.Join(absDir, rel))
+	// Ensure the cleaned path is still under absDir.
+	if cleaned != absDir && !strings.HasPrefix(cleaned, absDir+string(filepath.Separator)) {
+		return "", fmt.Errorf("manifest path %q escapes manifest directory", rel)
+	}
+	return cleaned, nil
+}
```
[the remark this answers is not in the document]
sthanikan2000: Addressed

backend/internal/config/config.go
```diff
@@ -11,9 +11,9 @@ import (
 	"github.com/OpenNSW/nsw/backend/internal/auth"
 	"github.com/OpenNSW/nsw/backend/internal/database"
 	"github.com/OpenNSW/nsw/backend/internal/temporal"
-	"github.com/OpenNSW/nsw/backend/internal/validation"
 	"github.com/OpenNSW/nsw/backend/pkg/blobsource"
 	"github.com/OpenNSW/nsw/backend/pkg/storage"
+	"github.com/OpenNSW/nsw/backend/pkg/validation"
```
ginaxu1, line 16: missing`getEnvOrDefault("BLOBSOURCE_GITHUB_MANIFEST_PATH" "manifest.json")` in `internal/config/config.go`

ginaxu1 (changes requested):

backend/pkg/blobsource/github.go
ginaxu1, line 254: above Line 248, to allow the http.Client's connection pool to reuse the underlying TCP/TLS connection via keep-alives, need to drain the response body before closing

```
_, _ = io.Copy(io.Discard, resp.Body)
```

backend/pkg/blobsource/github.go
```diff
@@ -54,18 +60,19 @@ type manifestData struct {
...
-	client   *http.Client
+	repo         string
+	ref          string
+	baseURL      string
+	manifestPath string // repo-relative path to the manifest file (default: "manifest.json")
+	interval     time.Duration
+	client       *http.Client
 
 	ctx    context.Context
 	cancel context.CancelFunc
```
ginaxu1, line 71: look into whether there's a background ticker Goroutine leak in `Close()`, and if we should add `sync.WaitGroup` to track and block until the background ticker loop has safely exited.

backend/pkg/blobsource/local_manifest.go
```diff
@@ -0,0 +1,81 @@
...
+// ".." or absolute paths. It also resolves symlinks so that a symlink pointing
+// outside the directory cannot bypass the check. absDir must already be the
+// real path (resolved via filepath.EvalSymlinks by the caller).
+func safeJoin(absDir, rel string) (string, error) {
+	if filepath.IsAbs(rel) {
+		return "", fmt.Errorf("manifest path %q must be relative", rel)
+	}
+	cleaned := filepath.Clean(filepath.Join(absDir, rel))
+	// Fast path: check before resolving symlinks.
+	if cleaned != absDir && !strings.HasPrefix(cleaned, absDir+string(filepath.Separator)) {
```
ginaxu1: nit: use `filepath.Rel` function (string-based directory containment checks are fragile across OS's) 
```go
relPath, err := filepath.Rel(absDir, cleaned)
if err != nil || strings.HasPrefix(relPath, "..") {
    return "", fmt.Errorf("manifest path %q escapes manifest directory", rel)
}
```

backend/pkg/blobsource/github.go
ginaxu1: `io.LimitReader` will silently cut the stream (if over a limit) and return EOF without throwing an error. 

better to rewrite line 250 to handle oversized payloads, e.g
```go
limitReader := io.LimitReader(resp.Body, maxResponseBytes+1)
body, err := io.ReadAll(limitReader)
if err != nil {
    return nil, fmt.Errorf("GET %s: read error: %w", requestURL, err)
}

// add another if check if int64(len(body)) > maxResponseBytes
return body, nil
```

ginaxu1 (changes requested):

backend/internal/config/config.go
sthanikan2000: Good Catch. Added.

backend/pkg/blobsource/github.go
sthanikan2000: Addressed

backend/pkg/blobsource/github.go
sthanikan2000: Addressed

backend/pkg/blobsource/local_manifest.go
sthanikan2000: Addressed

backend/pkg/blobsource/github.go
sthanikan2000: Addressed

PR #582, opened 2026-06-03
feat(deployment): add templates ConfigMap mounting for nsw-api + update DB seeds for fcau-health-cert

ginaxu1:
## Changes 

1. Core API Helm Chart & ConfigMap Mounting Support
- created `deployments/helm/nsw-api/templates/templates-cm.yaml` in the `nsw-api` Helm chart to support mounting workflow templates dynamically to `/app/configs/fcau` via ConfigMap
- updated `deployments/helm/nsw-api/templates/deployment.yaml` and `deployments/helm/nsw-api/values.yaml` to map the volumes and volume mounts for the templates ConfigMap
- bumped the `nsw-api` Helm chart version to `0.1.1` in `deployments/helm/nsw-api/Chart.yaml`

2. Agency Backend Helm Chart & Task Configs Mounting Support
- created `deployments/helm/nsw-agency-backend/templates/taskconfigs-cm.yaml` in the `nsw-agency-backend` Helm chart to support mounting task configurations dynamically to `/opt/nsw-agency/data/task-configs` via ConfigMap
- updated `deployments/helm/nsw-agency-backend/templates/deployment.yaml` to support mapping the volume and volume mounts for the task configs ConfigMap when `taskConfigs` is provided in the values
- bumped the `nsw-agency-backend` Helm chart version to `0.1.2` in `deployments/helm/nsw-agency-backend/Chart.yaml`

3. Database migration Seeds Refactor
- updated DB migration seeds in `015_fcau_workflow_seed.up.sql` and `015_fcau_workflow_seed.down.sql` to remove the hardcoded JSON definition of the `fcau-v1` workflow, seeding an empty JSON payload `{}` instead. This aligns with shifting the source of truth for workflow definitions from static DB seeds to dynamic templates loaded from disk (the mounted ConfigMap path `/app/configs/fcau`)
- Updated the HS code seeds mapping in the migration to map the testing HS code to the correct externalized workflow ID (`fcau-health-certificate-reg` instead of `fcau-v1`)

### GitOps Updates (nsw-gitops repo)
- Set up a temporary `local-charts` symlink pointing to `nsw/deployments/helm` to allow Helm to resolve local dependencies.
- Packaged the updated `nsw-api` Helm chart to `nsw-api-0.1.1.tgz` and the updated `nsw-agency-backend` Helm chart to `nsw-agency-backend-0.1.2.tgz`.
- Ran `helm dependency update envs/base/apps-umbrella` to rebuild the lockfile and update locked versions.
- Pushed `Chart.lock` and the packaged Helm archives directly to the `main` branch of the `nsw-gitops` repository


portals/apps/trader-app/src/main.tsx
```diff
@@ -44,6 +45,7 @@ initAppConfig()
             afterSignInUrl={APP_URL}
             afterSignOutUrl={APP_URL}
             scopes={IDP_SCOPES}
+            signInOptions={{ resource: IDP_RESOURCE }}
```
sthanikan2000, line 48: This is not the correct way to do this. Check the env config of trader-app in `nsw-srilanka`. You need to request correct scopes to get `NSW_API` as audience.
You need to add it to the helm values.

sthanikan2000 (commented):
For the migration of `nsw-backend` database we should use the migration defined in the `nsw-srilanka`. Can you check with it?

backend/internal/database/migrations/015_fcau_workflow_seed.up.sql
```diff
@@ -1,189 +1,18 @@
 INSERT INTO workflow_template_v2 (id, name, version, workflow_definition)
```
lokewate, line 1: any reason we are touch these files? we are deploying from nsw-srilanka right?

